PR #6, opened 2024-02-26
CICE grid generation from MOM supergrid 

ezhilsabareesh8:
This pull request addresses issues https://github.com/COSIMA/access-om3/issues/101 and https://github.com/COSIMA/access-om2/issues/280 by adding a Python script to generate the CICE6 grid from the MOM6 supergrid.

The script facilitates the creation of a CICE grid, resolving a previously encountered longitude mismatch issue by correctly positioning the U points in the NE corner.

### DIFF
--- a/grid_generation/Gen_CICE_grid.py
+++ b/grid_generation/Gen_CICE_grid.py
@@ -1,0 +1,99 @@
+"""
+Script: generate_cice_grid.py
+
+Description: 
+This script generates a CICE grid from the MOM super grid provided in the input NetCDF file.
+
+Contact: 
+Name: Ezhilsabareesh Kannadasan
+
+Usage:
+python generate_cice_grid.py <input_superGridPath> <output_file>
+- input_superGridPath: Path to the MOM super grid NetCDF file.
+- output_file: Path to the output CICE grid NetCDF file.
+"""
+import numpy as np
+import xarray as xr
+from netCDF4 import Dataset
+import sys
+
+def generate_cice_grid(in_superGridPath, output_file):
+    # Read input files
+    in_superGridFile = xr.open_dataset(in_superGridPath)
+
+    # Constants
+    ULAT = np.deg2rad(in_superGridFile['y'][1::2, 2::2])
+    ULON = np.deg2rad(in_superGridFile['x'][1::2, 2::2])
+    TLAT = np.deg2rad(in_superGridFile['y'][1::2, 1::2])
+    TLON = np.deg2rad(in_superGridFile['x'][1::2, 1::2])
+
+    # MPI
+    HTN = in_superGridFile['dx'] * 100.0  # convert to cm
+    HTN = HTN[1::2, ::2] + HTN[1::2, 1::2]
+    HTE = in_superGridFile['dy'] * 100.0  # convert to cm
+    HTE = HTE[::2, 1::2] + HTE[1::2, 1::2]
+
+    ANGLE = dtr * in_superGridFile['angle_dx'][1::2, 1::2]
+
+    # Close input files
+    in_superGridFile.close()
+
+    # Create a new NetCDF file
+    nc = Dataset(output_file, 'w', format='NETCDF4')
+
+    # Define dimensions
+    ny, nx = ULAT.shape
+    nc.createDimension('ny', ny)
+    nc.createDimension('nx', nx)
+
+    # Define variables
+    ulat = nc.createVariable('ulat', 'f8', ('ny', 'nx'))
+    ulon = nc.createVariable('ulon', 'f8', ('ny', 'nx'))
+    tlat = nc.createVariable('tlat', 'f8', ('ny', 'nx'))
+    tlon = nc.createVariable('tlon', 'f8', ('ny', 'nx'))
+    htn = nc.createVariable('htn', 'f8', ('ny', 'nx'))
+    hte = nc.createVariable('hte', 'f8', ('ny', 'nx'))
+    angle = nc.createVariable('angle', 'f8', ('ny', 'nx'))
+
+    # Add attributes
+    ulat.units = "radians"
+    ulat.title = "Latitude of U points"
+    ulon.units = "radians"
+    ulon.title = "Longitude of U points"
+    tlat.units = "radians"
+    tlat.title = "Latitude of T points"
+    tlon.units = "radians"
+    tlon.title = "Longitude of T points"
+    htn.units = "cm"
+    htn.title = "Width of T cells on North side."
+    hte.units = "cm"
+    hte.title = "Width of T cells on East side."
+    angle.units = "radians"
+    angle.title = "Rotation angle of U cells."
+
+    # Write data to variables
+    ulat[:] = ULAT
+    ulon[:] = ULON
+    tlat[:] = TLAT
+    tlon[:] = TLON
+    htn[:] = HTN
+    hte[:] = HTE
+    angle[:] = ANGLE
+
+    # Close the file
+    nc.close()
+
+    print("NetCDF file created successfully.")
+
+if __name__ == "__main__":
+    if len(sys.argv) != 3:
+        print("Usage: python script.py <input_superGridPath> <output_file>")
+        sys.exit(1)
+
+    input_superGridPath = sys.argv[1]
+    output_file = sys.argv[2]
+
+    # Constants
+    dtr = np.pi / 180.0  # Degree to Radian conversion
+
+    generate_cice_grid(input_superGridPath, output_file)
--- a/grid_generation/Gen_CICE_grid.py
+++ b/grid_generation/Gen_CICE_grid.py
@@ -22,7 +22,7 @@ def generate_cice_grid(in_superGridPath, output_file):
     in_superGridFile = xr.open_dataset(in_superGridPath)
 
     # Constants
-    ULAT = np.deg2rad(in_superGridFile['y'][1::2, 2::2])
+    ULAT = np.deg2rad(in_superGridFile['y'][2::2, 1::2])
     ULON = np.deg2rad(in_superGridFile['x'][1::2, 2::2])
     TLAT = np.deg2rad(in_superGridFile['y'][1::2, 1::2])
     TLON = np.deg2rad(in_superGridFile['x'][1::2, 1::2])
--- a/grid_generation/Gen_CICE_grid.py
+++ b/grid_generation/Gen_CICE_grid.py
@@ -65,7 +65,7 @@ def generate_cice_grid(in_superGridPath, output_file):
     tlon.units = "radians"
     tlon.title = "Longitude of T points"
     htn.units = "cm"
-    htn.title = "Width of T cells on North side."
+    htn.title = "Width of T cells on N side."
     hte.units = "cm"
     hte.title = "Width of T cells on East side."
     angle.units = "radians"
--- a/grid_generation/Gen_CICE_grid.py
+++ b/grid_generation/Gen_CICE_grid.py
@@ -1,5 +1,5 @@
 """
-Script: generate_cice_grid.py
+Script: Gen_CICE_grid.py
 
 Description: 
 This script generates a CICE grid from the MOM super grid provided in the input NetCDF file.
@@ -87,7 +87,7 @@ def generate_cice_grid(in_superGridPath, output_file):
 
 if __name__ == "__main__":
     if len(sys.argv) != 3:
-        print("Usage: python script.py <input_superGridPath> <output_file>")
+        print("Usage: python Gen_CICE_grid.py <input_superGridPath> <output_file>")
         sys.exit(1)
 
     input_superGridPath = sys.argv[1]
--- a/grid_generation/Gen_CICE_grid.py
+++ b/grid_generation/Gen_CICE_grid.py
@@ -65,7 +65,7 @@ def generate_cice_grid(in_superGridPath, output_file):
     HTE = in_superGridFile['dy'] * 100.0  # convert to cm
     HTE = HTE[::2, 2::2] + HTE[1::2, 2::2]
 
-    # The angle of rotation is a corner cell centres and applies to both t and u cells.
+    # The angle of rotation is at corner cell centres and applies to both t and u cells.
     ANGLE = np.deg2rad(in_superGridFile['angle_dx'][2::2, 2::2])
     ANGLET= np.deg2rad(in_superGridFile['angle_dx'][1::2, 1::2])
        
@@ -80,7 +80,7 @@ def generate_cice_grid(in_superGridPath, output_file):
     area_ext = np.append(area_ext[:], area_ext[-1:, :], axis=0)
 
     UAREA = area_ext[1::2, 1::2] + area_ext[2::2, 1::2] + \
-                    area_ext[2::2, 2::2] + area_ext[1::2, 2::2]
+                area_ext[2::2, 2::2] + area_ext[1::2, 2::2]
 
     # Close input files
     in_superGridFile.close()
@@ -94,16 +94,16 @@ def generate_cice_grid(in_superGridPath, output_file):
     nc.createDimension('nx', nx)
 
     # Define variables
-    ulat = nc.createVariable('ulat', 'f8', ('ny', 'nx'))
-    ulon = nc.createVariable('ulon', 'f8', ('ny', 'nx'))
-    tlat = nc.createVariable('tlat', 'f8', ('ny', 'nx'))
-    tlon = nc.createVariable('tlon', 'f8', ('ny', 'nx'))
-    htn = nc.createVariable('htn', 'f8', ('ny', 'nx'))
-    hte = nc.createVariable('hte', 'f8', ('ny', 'nx'))
-    angle = nc.createVariable('angle', 'f8', ('ny', 'nx'))
-    angleT = nc.createVariable('angleT', 'f8', ('ny', 'nx'))
-    tarea = nc.createVariable('tarea', 'f8', ('ny', 'nx'))
-    uarea = nc.createVariable('uarea', 'f8', ('ny', 'nx'))
+    ulat = nc.createVariable('ulat', 'f8', ('ny', 'nx'), compression='zlib', complevel=1)
+    ulon = nc.createVariable('ulon', 'f8', ('ny', 'nx'), compression='zlib', complevel=1)
+    tlat = nc.createVariable('tlat', 'f8', ('ny', 'nx'), compression='zlib', complevel=1)
+    tlon = nc.createVariable('tlon', 'f8', ('ny', 'nx'), compression='zlib', complevel=1)
+    htn = nc.createVariable('htn', 'f8', ('ny', 'nx'), compression='zlib', complevel=1)
+    hte = nc.createVariable('hte', 'f8', ('ny', 'nx'), compression='zlib', complevel=1)
+    angle = nc.createVariable('angle', 'f8', ('ny', 'nx'), compression='zlib', complevel=1)
+    angleT = nc.createVariable('angleT', 'f8', ('ny', 'nx'), compression='zlib', complevel=1)
+    tarea = nc.createVariable('tarea', 'f8', ('ny', 'nx'), compression='zlib', complevel=1)
+    uarea = nc.createVariable('uarea', 'f8', ('ny', 'nx'), compression='zlib', complevel=1)
 
     # Add attributes
     ulat.units = "radians"
--- a/grid_generation/Gen_CICE_grid.py
+++ b/grid_generation/Gen_CICE_grid.py
@@ -67,7 +67,7 @@ def generate_cice_grid(in_superGridPath, output_file):
     htn.units = "cm"
     htn.title = "Width of T cells on N side."
     hte.units = "cm"
-    hte.title = "Width of T cells on East side."
+    hte.title = "Width of T cells on E side."
     angle.units = "radians"
     angle.title = "Rotation angle of U cells."
 
--- a/grid_generation/Gen_CICE_grid.py
+++ b/grid_generation/Gen_CICE_grid.py
@@ -33,7 +33,26 @@ def generate_cice_grid(in_superGridPath, output_file):
     HTE = in_superGridFile['dy'] * 100.0  # convert to cm
     HTE = HTE[::2, 1::2] + HTE[1::2, 1::2]
 
-    ANGLE = np.deg2rad(in_superGridFile['angle_dx'][1::2, 1::2])
+    ANGLE = np.deg2rad(in_superGridFile['angle_dx'][2::2, 2::2])
+    ANGLET= np.deg2rad(in_superGridFile['angle_dx'][1::2, 1::2])
+       
+    # Area
+    AREA  = (in_superGridFile['area'])
+    TAREA = AREA[::2,::2] + AREA[1::2,1::2] + AREA[::2,1::2] +  AREA[1::2,::2]
+    
+    array_to_roll = AREA[2::2,2::2]
+    
+    # Roll the array along axis 0 and concatenate the last row as the first row
+    rolled_array_axis0 = np.concatenate((
+    np.roll(array_to_roll, -1, axis=0),
+    array_to_roll[-1:, :]), axis=0)
+
+    # Roll the rolled array along axis 1 and concatenate the last column as the first column
+    rolled_array = np.concatenate((
+    np.roll(rolled_array_axis0, -1, axis=1),
+    rolled_array_axis0[:, -1:]), axis=1)
+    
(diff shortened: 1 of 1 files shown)


grid_generation/Gen_CICE_grid.py
```diff
@@ -0,0 +1,96 @@
...
+from netCDF4 import Dataset
+import sys
+
+def generate_cice_grid(in_superGridPath, output_file):
+    # Read input files
+    in_superGridFile = xr.open_dataset(in_superGridPath)
+
+    # Constants
+    ULAT = np.deg2rad(in_superGridFile['y'][2::2, 1::2])
+    ULON = np.deg2rad(in_superGridFile['x'][1::2, 2::2])
```
anton-seaice, line 26: shouldn't this be 

```
ULAT = np.deg2rad(in_superGridFile['y'][2::2, 2::2])
ULON = np.deg2rad(in_superGridFile['x'][2::2, 2::2])
```

?

it might be clearer to write:
`np.deg2rad(in_superGridFile.y.sel(x=slice(start=2,step=2), y=slice(start=2, step=2)))`

?

aekiss: Agreed, but with `isel`, not `sel`

aekiss:
Yes, versioning info would be good too, e.g. https://github.com/COSIMA/om3-scripts/blob/main/mesh_generation/generate_mesh.py#L397-L413

aekiss:
It would be a good idea to compare the new `grid.nc` with the old one to check that it differs only in the ways we want it to.

There might be something in here that could be useful:
https://github.com/aekiss/notebooks/blob/master/check_updates.ipynb

grid_generation/Gen_CICE_grid.py
```diff
@@ -0,0 +1,96 @@
...
+    ulat.units = "radians"
+    ulat.title = "Latitude of U points"
+    ulon.units = "radians"
+    ulon.title = "Longitude of U points"
+    tlat.units = "radians"
+    tlat.title = "Latitude of T points"
+    tlon.units = "radians"
+    tlon.title = "Longitude of T points"
+    htn.units = "cm"
+    htn.title = "Width of T cells on North side."
```
aekiss, line 68:
```suggestion
    htn.title = "Width of T cells on N side."
```
(to match existing grid.nc)

grid_generation/Gen_CICE_grid.py
```diff
@@ -0,0 +1,96 @@
...
+    hte.units = "cm"
+    hte.title = "Width of T cells on East side."
```
aekiss, line 70:
```suggestion
    hte.title = "Width of T cells on E side."
```

aekiss:
This is missing the following from `/g/data/ik11/inputs/access-om2/input_20201102/cice_025deg/grid.nc` - are they needed in CICE? I noticed there are even more things in `/g/data/ik11/inputs/access-om2/input_20201102/cice_1deg/grid.nc` but they apparently aren't needed, so maybe the following are also unnecessary too...?
```
	double angleT(ny, nx) ;
		angleT:units = "radians" ;
		angleT:title = "Rotation angle of T cells." ;
	double tarea(ny, nx) ;
		tarea:units = "m^2" ;
		tarea:title = "Area of T cells." ;
	double uarea(ny, nx) ;
		uarea:units = "m^2" ;
		uarea:title = "Area of U cells." ;
```

anton-seaice:
Those three quantities are computer internally in CICE:

https://github.com/CICE-Consortium/CICE/blob/64177e3e5658711f5f19030ddfcc6927f1dbda5d/cicecore/cicedyn/infrastructure/ice_grid.F90#L952

We might want to think a little bit about whether computing the areas in a projection aware way improves accuracy. I guess the first check would be whether the old `uarea = HTN* HTE `

aekiss:
I wonder whether it would be better to make CICE read the MOM supergrid directly?

anton-seaice:
It looks like code was added to cice5 to read tarea, tlat and tlon from the cice grid file rather than calculate them:

https://github.com/COSIMA/cice5/blob/edcfa6f9c76ed05b63196ce4b5355fa5a8f4fe3a/source/ice_grid.F90#L662

```
#ifdef AusCOM
! we also let cice read in the following fields to avoid possible mismatch
! between cice and mom4.
! ( 8) TLAT
! ( 9) TLON
! (10) TAREA
! (11) UAREA
! (12) ANGLET
! the following fields are also available in grid.nc but not read in 'cos they
! are not critical 
!  LONT_BONDS
!  LATT_BONDS
!  LONU_BONDS
!  LATU_BONDS
#endif
```

anton-seaice:
There is something similar in the cmeps driver for CICE:

grid_generation/Gen_CICE_grid.py
```diff
@@ -0,0 +1,96 @@
+"""
+Script: generate_cice_grid.py
```
micaeljtoliveira, line 2: The name here does not match the actual file name. I would suggest renaming the file to `generate_cice_grid.py`.

grid_generation/Gen_CICE_grid.py
```diff
@@ -0,0 +1,96 @@
...
+    angle[:] = ANGLE
+
+    # Close the file
+    nc.close()
+
+    print("NetCDF file created successfully.")
+
+if __name__ == "__main__":
+    if len(sys.argv) != 3:
+        print("Usage: python script.py <input_superGridPath> <output_file>")
```
micaeljtoliveira, line 90: Also here I would be better to use the actual name of the script.

grid_generation/Gen_CICE_grid.py
ezhilsabareesh8: Thanks @micaeljtoliveira. I have updated the filenames in this commit  8cd8480

grid_generation/Gen_CICE_grid.py
ezhilsabareesh8: Thanks @micaeljtoliveira. I have updated the filenames in this commit  8cd8480

micaeljtoliveira:
@ezhilsabareesh8 Thanks for the changes! I think `generate_cice_grid.py.` was a better name, but the most important is that it's consistent.

anton-seaice:
@ezhilsabareesh8 - Can you test the new grid with OM2? 

I think you will need to add:

```
! ( 8) TLAT
! ( 9) TLON
! (10) TAREA
! (11) UAREA
! (12) ANGLET
```

for it to work in OM2.

Should we have two versions of this grid? One for OM2 and one for OM3 (with less variables?)



> I wonder whether it would be better to make CICE read the MOM supergrid directly?

Shall we split this into a different issue? Current OM3 implementation has CICE calculating area internally, which is problematic.

ezhilsabareesh8:
@anton-seaice  I can test OM2 with the new `grid.nc`. It seems we don't need two versions of `grid.nc`, as the additional data will likely be ignored by CICE.

The newly generated `grid.nc` contains the following data:
```
- `ulat`: Latitude of U points (radians)
- `ulon`: Longitude of U points (radians)
- `tlat`: Latitude of T points (radians)
- `tlon`: Longitude of T points (radians)
- `htn`: Width of T cells on the N side (cm)
- `hte`: Width of T cells on the E side (cm)
- `angle`: Rotation angle of U cells (radians) 
'''

Additionally, we may need to include `TAREA` and `UREA` for OM2.

**Comment by anton-seaice** [comment]
Oh yes - I was just thinking of two versions because its confusing if we have a grid.nc file loaded by OM3 that is then only using some of the variables.

**Comment by ezhilsabareesh8** [comment]
@aekiss, I've incorporated the TAREA and UAREA variables in the latest commit (09be9fb). However, I've noticed a discrepancy between the output of the newly generated `grid.nc` and the old OM2 `grid.nc` for a 1degree grid. I'm uncertain whether the issue lies with the CICE grid or if the wrapping is not implemented correctly in the new grid. Could you please have a look at the following output?

`(grid_ds['uarea'] - grid_old_ds['uarea']).plot()`

Screenshot 2024-03-01 at 10 58 30 am

**Comment by anton-seaice** [comment]
It looks like we need to something different at the fold. (Calculated from ugrid on top, old one below)

Screenshot 2024-03-01 at 2 22 47 pm

**Comment by ezhilsabareesh8** [comment]
There's a discrepancy in the TAREA values within the tripolar region, similar to Anton's observation [here]
Screenshot 2024-03-01 at 2 36 43 pm


```
difference = (grid_ds['tarea'] - grid_old_ds['tarea'])
difference.isel(ny=np.arange(240,300)).plot()
```

**Comment by aidanheerdegen** [comment]
Some good work going on in here! I'm just popping in to ask that while this is being done better we ensure the ice `grid.nc` files have better CF compliant metadata than has historically been the case.

Case in point, this 1 degree CICE grid `/g/data/vk83/experiments/inputs/access-om2/ice/grids/global.1deg/2020.05.30/grid.nc` (same file as `/g/data/ik11/inputs/access-om2/input_20200530/cice_1deg/grid.nc`).

`cf_xarray` is completely unable to intuit any CF coordinate information:
```
>>> import xarray, cf_xarray
>>> ds = xarray.open_dataset('/g/data/vk83/experiments/inputs/access-om2/ice/grids/global.1deg/2020.05.30/grid.nc')
>>> ds.cf
Coordinates:
             CF Axes:   X, Y, Z, T: n/a

      CF Coordinates:   longitude, latitude, vertical, time: n/a

       Cell Measures:   area, volume: n/a

      Standard Names:   n/a

              Bounds:   n/a

       Grid Mappings:   n/a

Data Variables:
       Cell Measures:   area, volume: n/a

      Standard Names:   n/a

              Bounds:   n/a

       Grid Mappings:   n/a
>>> 
```

Happy to move this to a separate issue if that is more appropriate.

**Comment by ezhilsabareesh8** [comment]
> Yes, versioning info would be good too, e.g. https://github.com/COSIMA/om3-scripts/blob/main/mesh_generation/generate_mesh.py#L397-L413

Thanks @aekiss, versioning information is added in this f7434b2 commit

**Comment by aekiss** [comment]
Just to clarify some terminology: the "fold" is the top of the numerical grid, physically corresponding to the line joining the two northern poles. See https://cice-consortium-cice.readthedocs.io/en/main/user_guide/ug_implementation.html#tripole-grids

[This difference](https://github.com/COSIMA/om3-scripts/pull/6#issuecomment-1972405759) is at the join between the lat-lon grid and the tripolar region (at 65°N). @micaeljtoliveira is this the oddity you mentioned to me last week?

**Comment by micaeljtoliveira** [comment]
@aekiss Yes, it seems so.

**Comment by aekiss** [comment]
The 1deg grid has meridional grid refinement near the equator, and coarsening elsewhere

download-1

(figure from [here](https://github.com/COSIMA/ACCESS-OM2-1-025-010deg-report/blob/master/figures/grid/grid.ipynb))

This may play a role in [these differences](https://github.com/COSIMA/om3-scripts/pull/6#issuecomment-1972387980) at mid-latitudes.

**Comment by aekiss** [comment]
re. the [differences at the tripole](https://github.com/COSIMA/om3-scripts/pull/6#issuecomment-1972454034) - not sure if this helps, but FYI we are using `grid_type = "tripole"` rather than `tripoleT`, so the northern poles and the fold between them all lie on U points rather than T points (see [here](https://cice-consortium-cice.readthedocs.io/en/main/user_guide/ug_implementation.html#tripole-grids)). This is to match the alignment in MOM5 - see [Elements of MOM](https://mom-ocean.github.io/assets/pdfs/MOM5_manual.pdf) sec 9.3 and fig 9.9 (copied below)
Screenshot 2024-03-04 at 2 48 43 pm

**Comment by aekiss** [comment]
Sec 9.4.3 of [Elements of MOM](https://mom-ocean.github.io/assets/pdfs/MOM5_manual.pdf) describes the special treatment of grid cell areas long the fold.

**Comment by aekiss** [comment]
Re. [the area mismatch](https://github.com/COSIMA/om3-scripts/pull/6#issuecomment-1972387980), I guess the original 1deg areas in the supergrid were calculated using [`update_cell_area`](https://github.com/mom-ocean/MOM5/blob/d7ba13a3f364ce130b6ad0ba813f01832cada7a2/src/preprocessing/generate_grids/ocean/hgrid.f90#L1123-L1140). Is the new code calculating areas the same way?

**Comment by navidcy** [comment]
Sorry to intrude in the discussion, somehow I get notifications for this. 

The are of a quadrilateral on the sphere is uniquely defined given the lon-lat coords of its four vertices. It doesn’t need to be approximated. So, regardless how it was done in the past, it’s a good idea to drop any unnecessary approximations. The sum of all grid areas should be precisely 4πR^2 (up to machine precision).

**Comment by navidcy** [comment]
See eg
https://github.com/COSIMA/regional-mom6/blob/b0e13c5f6a45eec9451ea2165f19ae0fe9176aed/regional_mom6/utils.py#L42

**Comment by navidcy** [comment]
There a way to avoid needing to ensure you give the vertices in counter-clockwise order. I can put more details when I am in front of a computer.

**Comment by anton-seaice** [comment]
> Sorry to intrude in the discussion, somehow I get notifications for this.
> 
> The area of a quadrilateral on the sphere is uniquely defined given the lon-lat coords of its four vertices. It doesn’t need to be approximated. So, regardless how it was done in the past, it’s a good idea to drop any unnecessary approximations. The sum of all grid areas should be precisely 4πR^2 (up to machine precision).

Thanks. We are trying to calculate the area of each cice grid cell by using the mom supergrid file. The areas in the mom supergrid seem to be calculated consistently with the way you are saying (https://github.com/ESMG/gridtools/blob/de0a18c1ce0807748aa70023300dfc415277bd4c/gridtools/spherical.py#L36).

This is a good thing for us to try though, if the areas calculated based on the lat/lons only are the same as the ones derived from the mom supergrid, that would be a good validation of the method.

**Comment by anton-seaice** [comment]
> Re. [the area mismatch](https://github.com/COSIMA/om3-scripts/pull/6#issuecomment-1972387980), I guess the original 1deg areas in the supergrid were calculated using [`update_cell_area`](https://github.com/mom-ocean/MOM5/blob/d7ba13a3f364ce130b6ad0ba813f01832cada7a2/src/preprocessing/generate_grids/ocean/hgrid.f90#L1123-L1140). Is the new code calculating areas the same way?

All we are doing is summing the area of the four relevant cells from the supergrid to make the cice grid, I can't see anything inconsistent with the update_cell_area in that? 

This is similar to what Navid is saving I think - we could use this method to calculate area based on lat/lon only, and not from the supergrid.

**Comment by anton-seaice** [comment]
> The 1deg grid has meridional grid refinement near the equator, and coarsening elsewhere
> 
> download-1
> 
> (figure from [here](https://github.com/COSIMA/ACCESS-OM2-1-025-010deg-report/blob/master/figures/grid/grid.ipynb))
> 
> This may play a role in [these differences](https://github.com/COSIMA/om3-scripts/pull/6#issuecomment-1972387980) at mid-latitudes.

Does this mean the coordinates move away from regular 1 degree spacing?

**Comment by anton-seaice** [comment]
The patterning found above in t-area seems to only be there in the 1-degree grid, so it might be an artifact of the Meridional refinement and the grids getting out of sync at some point. Its not there in the 0.25 degree grid, so maybe we can ignore it.

The difference in u-area, is still strange. In our current grid at the join:

image

and in the calculating from the mom supergrid at the join:

image

I think the second one is correct. (Amusingly the files in '/g/data/ik11/grids/' also have one of each option)

This figure kind of demonstrates it (viewed straight down on the north pole):

image

u-area is the sum of the areas of a quarter of each grid cell which surrounds each u-point (i.e. NE corner point). The areas north of the join (The thick line) are smaller than those south of the join with a notable change. So the u-areas at the join are the average of the smaller and larger cell areas which exist either side of the join, and therefore create the 'step' shown in the second figure.




As an aside, I stumbled across this code which Nic wrote and does what we want:

https://github.com/COSIMA/esmgrids/blob/130c8d00774c624c196fc2ca15f5cad78c4cdbc1/esmgrids/mom_grid.py

**Comment by aekiss** [comment]
> Does this mean the coordinates move away from regular 1 degree spacing?

yes, the meridional (y) spacing is refined to 1/3° near the equator on the 1° grid.

Note that the y spacing is not regular at most latitudes at 0.25° or 0.1° either, because a Mercator grid is used between 65S and 65N to keep the cell aspect ratio equal to 1. See https://github.com/COSIMA/ACCESS-OM2-1-025-010deg-report/blob/master/figures/grid/grid.ipynb

> p.s. Its also odd that the patterning below 65N is there in the uarea calc and not the tarea calc.

yes, indeed

**Comment by aekiss** [comment]
I think I agree with [your analysis](https://github.com/COSIMA/om3-scripts/pull/6#issuecomment-1980124729) @anton-seaice, so long as the 65°N junction between the Mercator and tripolar grids falls on T points - is that the case?

**Comment by anton-seaice** [comment]
> I think I agree with [your analysis](https://github.com/COSIMA/om3-scripts/pull/6#issuecomment-1980124729) @anton-seaice, so long as the 65°N junction between the Mercator and tripolar grids falls on U points - is that the case?

Good points - I check - all three resolutions have the join at u-points

**Comment by ezhilsabareesh8** [comment]
The angle ANGLET in the current implementation is calculated directly from the super grid as  [Refer here](https://github.com/COSIMA/om3-scripts/blob/90f1e352f116bbda88632c7a4f667629f291c2f0/grid_generation/Gen_CICE_grid.py#L69-L70)


`ANGLE = np.deg2rad(in_superGridFile['angle_dx'][2::2, 2::2]) ;
    ANGLET= np.deg2rad(in_superGridFile['angle_dx'][1::2, 1::2])`

which is consistent with the OM2 `grid.nc` file. However, it appears that this differs from the weighted average approach employed in CICE6 [here](https://github.com/CICE-Consortium/CICE/blob/740f2a65e055e37bebc81aa25675ebc2a65d735d/cicecore/cicedyn/infrastructure/ice_grid.F90#L617-L632).

Should we consider adapting the angleT calculation method to utilize the weighted average approach used in CICE6?

File: grid_generation/Gen_CICE_grid.py
Comment (ezhilsabareesh8) [inline grid_generation/Gen_CICE_grid.py:1]: The discrepancy in wrapping is fixed by changing the wrapping similar to `esmgrids` [refer here](https://github.com/COSIMA/esmgrids/blob/514b560adf7b9094081408da49ecb3e19ef0353b/esmgrids/mom_grid.py#L102-L110)

File: grid_generation/Gen_CICE_grid.py [hunk 6]
```diff
@@ -0,0 +1,167 @@
...
+    in_superGridFile = xr.open_dataset(in_superGridPath)
+
+    # Constants
+    ULAT = np.deg2rad(in_superGridFile['y'][2::2, 2::2])
+    ULON = np.deg2rad(in_superGridFile['x'][2::2, 2::2])
+    TLAT = np.deg2rad(in_superGridFile['y'][1::2, 1::2])
+    TLON = np.deg2rad(in_superGridFile['x'][1::2, 1::2])
+
+    HTN = in_superGridFile['dx'] * 100.0  # convert to cm
+    HTN = HTN[1::2, ::2] + HTN[1::2, 1::2]
```
anton-seaice, line 64: Screenshot 2024-03-07 at 4 15 14 pm


In the mom_grid, dx has units ('nyp','nx'). These correspond to the edges in y and the middle in x. So, the y coordinate (i.e. rows) should be 2::2

grid_generation/Gen_CICE_grid.py
```diff
@@ -60,32 +60,27 @@ def generate_cice_grid(in_superGridPath, output_file):
...
-    rolled_array = np.concatenate((
-    np.roll(rolled_array_axis0, -1, axis=1),
-    rolled_array_axis0[:, -1:]), axis=1)
-    
-    UAREA = AREA[1::2,1::2] + np.concatenate((np.roll(AREA[1::2, 2::2], -1, axis=1), AREA[1::2, 2::2][:, :1]), axis=1) + np.concatenate((np.roll(AREA[2::2, 1::2], -1, axis=0), AREA[2::2, 1::2][:1, :]), axis=0) + rolled_array
+    # UAREA need to wrap around the globe. Copy ocn_area and
+    # add an extra column at the end. Also u-cells cross the
+    # tri-polar fold so add an extra row at the top.
+    area_ext = np.append(AREA[:], AREA[:, 0:1], axis=1)
+    area_ext = np.append(area_ext[:], area_ext[-1:, :], axis=0)
```
anton-seaice, line 80: I think this line only works because the top row is symmetrical? It doesn't actual seem to fold it... ?

anton-seaice:
I did a plot with every var and every resolution (left to right: 1deg, 0.25deg, 0.1deg). I masked difference which were less than 2e-6

- Need to investigate: Angle,  0.1 ULAT/ULON
- UAREA doesn't match well at any resolution, which is problematic.


I think we should merge this PR soon. We still have to add CF metadata and test with OM2, but these could be new issues to stop this PR totally blowing out? We also need to decide if we are sticking with np syntax (e.g. [2::2,1::2] or moving to written xarray style (e.g. `ds.coarsen` and `ds.isel(x=slice(...))`)


image

anton-seaice:
I turned on compression:

Before

```
-rw-r--r-- 1 as2285 tm70 742M Mar  7 16:27 grid_01deg.nc
-rw-r--r-- 1 as2285 tm70 119M Mar  7 16:27 grid_025deg.nc
-rw-r--r-- 1 as2285 tm70 8.3M Mar  7 16:27 grid_1deg.nc
```

After:

```
-rw-r--r-- 1 as2285 tm70  123M Mar  8 10:47 grid_01deg.nc
-rw-r--r-- 1 as2285 tm70   15M Mar  8 10:46 grid_025deg.nc
-rw-r--r-- 1 as2285 tm70 1019K Mar  8 10:46 grid_1deg.nc
```

I left chunking at the default sizes, which gives 0.1 deg chunks of  900, 1200 (y,x) but the other resolutions are full size

grid_generation/Gen_CICE_grid.py
```diff
@@ -0,0 +1,167 @@
...
+    if is_git_repo():
+        git_url, git_hash, rel_path = git_info()
+        nc.inputfile = f"{in_superGridPath}"
+        nc.timeGenerated = f"{datetime.now()}"
+        nc.created_by = f"{os.environ.get('USER')}"
+        nc.history = f"Created using commit {git_hash} of {git_url}"
+    else:
+        nc.inputfile = f"{in_superGridPath}"
+        nc.timeGenerated = f"{datetime.now()}"
+        nc.created_by = f"{os.environ.get('USER')}"
```
anton-seaice, line 140:
```
nc.inputfile = f"{in_superGridPath}"
nc.timeGenerated = f"{datetime.now()}"
nc.created_by = f"{os.environ.get('USER')}"
```

These three lines can go outside the if/else as they are the same in both cases.

Can we add a hash for the input file? Similar to the payu manifests md5 hash?

anton-seaice:
It looks like the old 0.1 degree grid was wrong too.

The x & y are offset by 1 in both lat and lon, i.e. ulat was in the SW corner, not the NE corner.

Screenshot 2024-03-08 at 2 53 26 pm

The differences in 'angle' above are related to the same problem with the wrong corner being referenced in the old grid. (See plots [here ](https://github.com/anton-seaice/sandbox/blob/1ce2d8400bfb685766b7f86b8a97f558abf2e707/cice_grid_valiation.ipynb))

I guess we need a new issue in the 0.1 degree repo to test the new grid as well.

anton-seaice:
I compared every resolution & every var between the version from this script, and the version in '/g/data/ik11/inputs/access-om2/input_20200530/'

- ulat/ulon/angle for 0.25 degree resolution has been updated to be referencing the correct corner now
- angleT in 0.01 degree looks wrong, as [Ezhil noted](https://github.com/COSIMA/om3-scripts/pull/6#issuecomment-1982121114)
- as discussed, uarea has changed at 1deg/0/25 deg resolution


image

anton-seaice:
I suggest - to remove duplication in code and make the result more extendable we close this PR and re-write this using https://github.com/COSIMA/esmgrids.

The script can be based on https://github.com/COSIMA/access-om2/blob/master/tools/make_cice_grid.py but we need add version control, fix AngleT in esmgrids and add CF metadata.

Using esmgrids makes this easier to extend in the future (e.g. to make the ocean_grid file for analysis).

anton-seaice:
> I suggest - to remove duplication in code and make the result more extendable we close this PR and re-write this using https://github.com/COSIMA/esmgrids.
> 
> The script can be based on https://github.com/COSIMA/access-om2/blob/master/tools/make_cice_grid.py but we need add version control, fix AngleT in esmgrids and add CF metadata.
> 
> Using esmgrids makes this easier to extend in the future (e.g. to make the ocean_grid file for analysis).

@aekiss  thoughts ?

anton-seaice:
Thanks for all your work on this Ezhil. I've taken our lessons and re-worked using esmgrids in a [new PR](https://github.com/COSIMA/om3-scripts/pull/8) :)

PR #10, opened 2024-04-05
`generate_mesh.py`: improve generation of history metadata

### LINKED ISSUE #9: Improve metadata added by `generate_mesh.py`
For better provenance, we should:

- make sure paths are written as absolute paths
- include hashes of the input files
- fix logic so that git info is still added when command is run from outside the repo

dougiesquire:
Closes #9

### DIFF
--- a/mesh_generation/generate_mesh.py
+++ b/mesh_generation/generate_mesh.py
@@ -23,6 +23,7 @@
 import os
 import io
 import hashlib
+import warnings
 import subprocess
 from datetime import datetime
 
@@ -58,6 +59,28 @@ def get_git_url(file):
     ).decode("ascii").strip()
 
     return f"{url}/blob/{hash}{rel_path}"
+
+def git_status(file):
+    """
+    Return the git status of the file. Returns:
+    - "unstaged" if the file has unstaged changes
+    - "uncommitted" if the file has uncommited changes,
+    - "unpushed" if the repo has unpushed commits
+    - None otherwise
+    """
+    dirname = os.path.dirname(file)
+    status = subprocess.check_output(
+        ["git", "-C", dirname, "status", file]
+    ).decode("ascii").strip()
+
+    if "Changes not staged for commit" in status:
+        return "unstaged"
+    elif "Changes to be committed" in status:
+        return "uncommitted"
+    elif "Your branch is ahead" in status:
+        return "unpushed"
+    else:
+        return None
 
 def md5sum(path):
     """
@@ -428,6 +451,11 @@ def main():
     git_url = get_git_url(this_file)
 
     if git_url:
+        status = git_status(this_file)
+        if status in ["unstaged", "uncommitted"]:
+            warnings.warn(f"{this_file} contains uncommitted changes! Commit and push your changes before generating any production output.")
+        if status == "unpushed":
+            warnings.warn(f"There are commits that are not pushed! Push your changes before generating any production output.")
         prepend = f"Created using {git_url}: "
     else:
         prepend = f"Created using {this_file}: "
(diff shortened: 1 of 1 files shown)


mesh_generation/generate_mesh.py
```diff
@@ -29,35 +31,46 @@
...
-    ).decode('ascii').strip()
-    rel_path = os.path.relpath(__file__, top_level_dir)
+        ["git", "-C", dirname, "rev-parse", "--show-toplevel"]
+    ).decode("ascii").strip()
+    rel_path = file.removeprefix(top_level_dir)
+
     hash = subprocess.check_output(
-        ['git', 'rev-parse', 'HEAD']
-    ).decode('ascii').strip()
+        ["git", "-C", dirname, "rev-parse", "HEAD"]
```
aekiss, line 57: could use `--short` if you'd rather a shorter git hash, though full hash is safer...

aekiss (approved):
Thanks @dougiesquire, looks great.

Since we will want to record this sort of detailed provenance for all our generated files, could/should these parts be split off as a general-purpose tool somehow?

mesh_generation/generate_mesh.py
```diff
@@ -29,35 +31,46 @@
...
+    ).decode("ascii").strip()
 
-    return url, rel_path, hash
+    return f"{url}/blob/{hash}{rel_path}"
```
aekiss, line 60: or maybe
```suggestion
    return f"{url}/tree/{hash}{rel_path}"
```
? I don't really know the difference, so feel free to ignore...

aekiss:
What if the current dir has uncommitted changes, or the changes haven't been pushed to the remote? Should the script abort and request a git commit and push so the hash will be correct?

FYI I split off https://github.com/COSIMA/make_025deg_topo/blob/master/finalise.sh as a separate script to allow experimental runs of the script, and only committing and recording provenance as a separate final step.

mesh_generation/generate_mesh.py
dougiesquire: > 

Neither did I. Apparently [`blob` is for files, `tree` is for directories](https://stackoverflow.com/questions/39400848/in-github-urls-what-is-the-difference-between-a-tree-and-a-blob). But the browser redirects to whichever is correct, so it doesn't really matter. I'll leave as `blob` since this is a file.

dougiesquire:
> What if the current dir has uncommitted changes, or the changes haven't been pushed to the remote? Should the script abort and request a git commit and push so the hash will be correct?
> 
> FYI I split off https://github.com/COSIMA/make_025deg_topo/blob/master/finalise.sh as a separate script to allow experimental runs of the script, and only committing and recording provenance as a separate final step.

What about issuing a warning? Then one can still do experimental runs, but hopefully it will remind people to push before generating production versions?

dougiesquire:
> Since we will want to record this sort of detailed provenance for all our generated files, could/should these parts be split off as a general-purpose tool somehow?

Yes definitely! I think a good options is to develop [`esmgrids`](https://github.com/COSIMA/esmgrids) to become our package for working with OM3 grids (conversion, generation etc). The mesh generation functionality could be moved there, as could the provenance metadata utils.

anton-seaice:
I didn't review in detail but maybe we should add a README, or a comment at the top of the file about the git commits. i.e. 

_The hash of the git commit is added to the metadata of the mesh netcdf files this script produces to uniquely identify the configuration used to generate the mesh file. To produce mesh files for sharing, ensure you are using a version of this script which is committed to git. For mesh files intended for released configuration, use the latest version checked in to the main branch of the git repository._

aekiss:
also needs to be pushed to github:

> _The hash of the git commit is added to the metadata of the mesh netcdf files this script produces to uniquely identify the configuration used to generate the mesh file. To produce mesh files for sharing, ensure you are using a version of this script which is **committed** to git and **pushed to GitHub**. For mesh files intended for a released configuration, use the latest version checked in to the main branch of the git repository._

PR #12, opened 2024-07-15
move md5sum and git to tools module

anton-seaice:
Does this seem like an ok way to generalise the metadata tools ?

If this looks ok, I can add copyright information etc .


anton-seaice:
Ill wait for you to confirm, but I guess moving everything in the tools folder to a `utils.py` in the root directory is the best way forward?

also i need some help with naming these functions !!

tools/md5sum.py
```diff
@@ -0,0 +1,16 @@
+import io
+import hashlib
+
+
+def md5sum(path):
```
dougiesquire, line 5: There's a name clash here with the sub-module name, but this will go away once we move everything into a single module

tools/git.py
```diff
@@ -0,0 +1,74 @@
...
+    if "Changes not staged for commit" in status:
+        return "unstaged"
+    elif "Changes to be committed" in status:
+        return "uncommitted"
+    elif "Your branch is ahead" in status:
+        return "unpushed"
+    else:
+        return None
+    
+def get_created_str(file) :
```
dougiesquire, line 56: I think we should also pass the run cmd to this to make it clear that this should always be included. Then the function could be something like:

```python
def get_provenance_metadata(file, runcmd):
    ...
```

mesh_generation/generate_mesh.py
```diff
@@ -27,79 +27,19 @@
...
-    From https://stackoverflow.com/a/40961519
-    """
-    length = io.DEFAULT_BUFFER_SIZE
-    md5 = hashlib.md5()
-    with io.open(path, mode="rb") as fd:
-        for chunk in iter(lambda: fd.read(length), b''):
-            md5.update(chunk)
-    return md5.hexdigest()
+from tools.git import get_created_str
+from tools.md5sum import *
```
dougiesquire, line 42: Could we do this with a relative import instead to avoid having to add to `PYTHONPATH`?

dougiesquire (commented):
Thanks for this @anton-seaice. Yeah I think it's simplest/cleanest just to put all the common functions in a single `.py` file in the base dir. If we are going to modify `PYTHONPATH` it's probably a good idea to name this file something unique to avoid namespace clashes, e.g. `om3_scripts_common.py`

mesh_generation/generate_mesh.py
dougiesquire: Yeah, I think we'd need to add some `__init__.py`s around the place and that probably won't help with clarity... Let's stick with adding to `PYTHONPATH` but name the new module something we can be pretty sure is unique

anton-seaice:
I ran this without any warnings: 

 ```
module use /g/data/hh5/public/modules ; module load conda/analysis
python3 mesh_generation/generate_mesh.py --grid-type mom --grid-filename /g/data/ik11/inputs/access-om2/input_20230515_025deg_topog/mom_025deg/ocean_hgrid.nc --mesh-filename test.nc
ncdump -hs test.nc 
```

The global attributes contain this, (which looks correct).

`
:history = "Created using https://github.com/COSIMA/om3-scripts/blob/a72e66648fbca5ae1adbe7fd15e6aaa40ff6abd1/mesh_generation/generate_mesh.py: python3 generate_mesh.py --grid-type=mom --grid-filename=/g/data/ik11/inputs/access-om2/input_20230515_025deg_topog/mom_025deg/ocean_hgrid.nc --mesh-filename=/g/data/tm70/as2285/om3-scripts/test.nc" ;
`

anton-seaice:
This is complete @dougiesquire

PR #13, opened 2024-07-16
OM3 user scripts

anton-seaice:
To use, add these to `config.yaml`:

```
userscripts:
    setup: /usr/bin/bash /g/data/tm70/as2285/om3-scripts/payu_config/setup.sh
    archive: /usr/bin/bash /g/data/tm70/as2285/om3-scripts/payu_config/archive.sh

modules:
    use: 
        - /g/data/hh5/public/modules
    load: 
        - conda/analysis
```

For reasons I don't I understand, putting module load statements within the scripts don't work.

Questions (@dougiesquire) :

- <s>is `conda/access-med-0.7` the right place to load intake from ?</s>
- is 48 workers the right number for building the intake datastore ? (should we use more at higher resolution?)
- should we write tests for this ?
- at the end of bash script "concat_ice_daily" we could add the git version and hash of the script used. Should we do this in python using the functions from other scripts? Or make another version of those functions in bash?


anton-seaice:
I intermittently get this error, but the job appears to succeed still:

2024-07-16 14:54:28,633 - distributed.worker - ERROR - Failed to communicate with scheduler during heartbeat.
Traceback (most recent call last):
  File "/g/data/xp65/public/apps/med_conda/envs/access-med-0.7/lib/python3.11/site-packages/distributed/comm/tcp.py", line 225, in read
    frames_nosplit_nbytes_bin = await stream.read_bytes(fmt_size)
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tornado.iostream.StreamClosedError: Stream is closed

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/g/data/xp65/public/apps/med_conda/envs/access-med-0.7/lib/python3.11/site-packages/distributed/worker.py", line 1252, in heartbeat
    response = await retry_operation(
               ^^^^^^^^^^^^^^^^^^^^^^
  File "/g/data/xp65/public/apps/med_conda/envs/access-med-0.7/lib/python3.11/site-packages/distributed/utils_comm.py", line 455, in retry_operation
    return await retry(
           ^^^^^^^^^^^^
  File "/g/data/xp65/public/apps/med_conda/envs/access-med-0.7/lib/python3.11/site-packages/distributed/utils_comm.py", line 434, in retry
    return await coro()
           ^^^^^^^^^^^^
  File "/g/data/xp65/public/apps/med_conda/envs/access-med-0.7/lib/python3.11/site-packages/distributed/core.py", line 1395, in send_recv_from_rpc
    return await send_recv(comm=comm, op=key, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/g/data/xp65/public/apps/med_conda/envs/access-med-0.7/lib/python3.11/site-packages/distributed/core.py", line 1154, in send_recv
    response = await comm.read(deserializers=deserializers)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/g/data/xp65/public/apps/med_conda/envs/access-med-0.7/lib/python3.11/site-packages/distributed/comm/tcp.py", line 236, in read
    convert_stream_closed_error(self, e)
  File "/g/data/xp65/public/apps/med_conda/envs/access-med-0.7/lib/python3.11/site-packages/distributed/comm/tcp.py", line 142, in convert_stream_closed_error
    raise CommClosedError(f"in {obj}: {exc}") from exc
distributed.comm.core.CommClosedError: in <TCP (closed) ConnectionPool.heartbeat_worker local=tcp://127.0.0.1:58844 remote=tcp://127.0.0.1:36209>: Stream is closed

dougiesquire:
Thanks @anton-seaice!

>  * is `conda/access-med-0.7` the right place to load intake from ?

I'd probably use `hh5`

> * is 48 workers the right number for building the intake datastore ? (should we use more at higher resolution?)

The datastore builders don't use dask for parallelization, so you shouldn't initialise a dask cluster. The number of cpus is automatically detected.

>  * should we write tests for this ?

Ideally we would have tests for everything. If you have the energy and inclination to set up tests, please do!

> * at the end of bash script "concat_ice_daily" we could add the git version and hash of the script used. Should we do this in python using the functions from other scripts? Or make another version of those functions in bash?

Add the git version and hash to what? The cice netcdf files? I think having a single version of the git/has functions is preferable

dougiesquire:
> Oh right! Is there an issue if there too many cpus (i.e. thousands)

No idea, but I guess we'll find out. If so, we can pass the number of cpus to use.

> But could add the github script as well

My opinion is that what's in there already is sufficient. It completely describes what was done to that file in the script

payu/archive_scripts/archive_cice_restarts.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/bash -i
+# clean up cice_restarts.sh
+
+latest_o=$(ls -drv archive/output*[0-9] | head -1)
+
+if [ -f $latest_o/access-om3.cice.r.* ]
```
anton-seaice, line 6: Do we want to update this to allow for multiple restarts ?

dougiesquire: When would that be needed? Are we likely to need it?

anton-seaice: Currently, if configured to save extra restarts (i.e. those without "restarting" the model), this line will just fail, and not remove the restart file which was used for to start the current model run.

dougiesquire: If it's easy to change to handle that case robustly, then I guess there's no reason not to

anton-seaice:
@dougiesquire This is ready to go. Let me know if you want to keep `archive.py` and `setup.py`. In future payu versions we should be able to call the individual scripts directly.

anton-seaice:
Also, if we want to change the cice output file names in post processing, we could do that now.

anton-seaice:
p.s.s. We can split into several PRs if that is better

payu/archive_scripts/build_intake_ds.py
```diff
@@ -0,0 +1,57 @@
...
+UUID_FIELD = "experiment_uuid"
+
+builder = builders.AccessOm3Builder(path="archive/")
+
+builder.build()
+
+# Log invalid assets
+builder.invalid_assets
+
+# Get experiment uuid
```
anton-seaice, line 34: I could probably import a fucntion to do this from a payu submodule? That probably makes more sense ?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,51 @@
...
+  formatting:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Code formatting
+      uses: psf/black@stable
+      with:
+        options: "--check --verbose --diff"
+
+  build:
```
dougiesquire, line 24:
```suggestion
  test:
```
Take or leave, but this might be more appropriate for what is being done in this task

.pre-commit-config.yaml
```diff
@@ -0,0 +1,10 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+  - repo: local
+    hooks:
+      - id: black
+        name: black
+        entry: black
+        language: system
+        types: [python]
```
dougiesquire, line 10: It would be good to add info to the root README encouraging people to use this and explaining how to set it up:

- `pip/conda install pre-commit black`
- `pre-commit install`

payu/archive_scripts/build_intake_ds.py
dougiesquire: Something like (untested):

```python
from payu.metadata import Metadata

metadata = Metadata("archive/")
uuid = metadata.uuid
```

You'll still need to handle the case where `UUID_FIELD` does not exist. Currently you're setting `uuid` to `None` in that case, which will fail at line 54.

dougiesquire (commented):
Looks great @anton-seaice! Just doing a test model run now. In the meantime, a few comments. Could you also add a README to the `payu` directory with details about how to use the scripts - i.e. basically what you have in the PR description, but I guess using the `hh5` conda env rather than `xp65`

payu/archive_scripts/concat_ice_daily.sh
```diff
@@ -0,0 +1,69 @@
...
+    echo "ncrcat not available, trying module load nco"
+    module load nco
+fi
+
+for f in $out_dir/access-om3.cice.h.????-??-01.nc ; do
+   #if the 1st and the 28th of that month exists, then assume its a whole month and concatenate
+   if [ -f $f ] && [ -f ${f/-01.nc/-28.nc} ]; then 
+
+      output_f=${f/-01.nc/.nc} #remove day in date string
+      output_f=${output_f/.cice.h./.cice.h.day.} #add word day to filename
```
dougiesquire, line 59: Thinking about this more, I think `daily`, `monthly` etc may be better for clarity to users

payu/archive_scripts/build_intake_ds.py
```diff
@@ -0,0 +1,57 @@
...
+# follows https://github.com/payu-org/payu/blob/ef55e93fe23fcde19024479c0dc4112dcdf6603f/payu/metadata.py#L90
+metadata_filename = Path(METADATA_FILENAME)
+if metadata_filename.exists():
+    metadata = CommentedMap()
+    metadata = YAML().load(metadata_filename)
+    uuid = metadata.get(UUID_FIELD, None)
+else:
+    warn(f"{METADATA_FILENAME} not found in archive folder")
```
dougiesquire, line 42: `warn` is not defined

dougiesquire (changes requested):
The model test run was mostly successful, except that the intake-esm datastore did not build because I've turned off the metadata stuff in Payu. I think we still want to build a datastore in this case as just call it `intake_esm_datastore` without the uuid appended.

dougiesquire:
Apologies for the multiple reviews - back to you now @anton-seaice

payu/archive_scripts/build_intake_ds.py
anton-seaice: Is the metadata worth including in the filename? I would still be in the description already.

Its easier to generalise opening $run_dir/intake_ds.json than $run_dir/intake_ds_#######.json ?

dougiesquire: Yeah true - let's just remove it. Could we call it `intake_esm_ds` though?

payu/archive_scripts/concat_ice_daily.sh
dougiesquire: It think we're heading towards `1day` - see [here](https://github.com/COSIMA/access-om3/issues/190)

anton-seaice:
@dougiesquire I think this is finished now

test/test_payu_conf/test_concat_ice_daily.py
```diff
@@ -0,0 +1,138 @@
...
+                run_str,
+                "-d",
+                output_dir,
+            ]
+        )
+        expected_months = pd.date_range("2010-01-01", freq="ME", periods=nmonths + 1)
+
+    # valid output filenames
+    monthly_paths = [
+        f"{output_dir}/access-om3.cice.h.1day.{str(t)[0:7]}.nc" for t in expected_months
```
dougiesquire, line 103: See above comment

payu_config/archive_scripts/build_intake_ds.py
```diff
@@ -0,0 +1,60 @@
...
+# Copyright 2024 ACCESS-NRI and contributors. See the top-level COPYRIGHT file for details.
+# SPDX-License-Identifier: Apache-2.0
+# modules:
+# use:
+#     - /g/data/hh5/public/modules
+# load:
+#     - conda/analysis
+
+from access_nri_intake.source import builders
+from payu.metadata import Metadata
```
dougiesquire, line 11: This doesn't work with `hh5` Payu and for some reason with the following in my `config.yaml` the script couldn't find `access_nri_intake` 

```
use:
    - /g/data/hh5/public/modules
    - /g/data/vk83/modules
load:
    - conda/analysis
    - payu
```

dougiesquire (changes requested):
The datastore building doesn't work for me (see below). Maybe it's safest to parse the uuid manually after all, rather than having Payu as a dependency?

payu_config/archive_scripts/build_intake_ds.py
dougiesquire: These two environments cannot be loaded simultaneously

payu_config/archive_scripts/concat_ice_daily.sh
```diff
@@ -0,0 +1,69 @@
...
+    echo "ncrcat not available, trying module load nco"
+    module load nco
+fi
+
+for f in $out_dir/access-om3.cice.h.????-??-01.nc ; do
+   #if the 1st and the 28th of that month exists, then assume its a whole month and concatenate
+   if [ -f $f ] && [ -f ${f/-01.nc/-28.nc} ]; then 
+
+      output_f=${f/-01.nc/.nc} #remove day in date string
+      output_f=${output_f/.cice.h./.cice.h.1day.} #add word day to filename
```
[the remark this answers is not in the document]
dougiesquire: Oh, and add `_mean`, e.g.

`access-om3.cice.1day_mean_1900_01.nc`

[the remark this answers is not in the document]
dougiesquire: Ugh, CICE uses dashes in the timestamp. I guess this is okay for now:

`access-om3.cice.1day_mean_1900-01.nc`

payu_config/archive_scripts/build_intake_ds.py
anton-seaice: At some point I think we want to add the intake catalog to vk83 to make this easier, but to progress this ill remove the payu dependency

dougiesquire:
I added a README, I hope you don't mind. This runs for me now. Just the CICE filename to update and then I think we're good to go!

payu_config/archive_scripts/concat_ice_daily.sh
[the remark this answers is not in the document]
dougiesquire: See https://github.com/COSIMA/access-om3/issues/190#issuecomment-2261753828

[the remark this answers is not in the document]
anton-seaice: I will remove the renaming and we can set the name directly from the model / its config

dougiesquire (approved):
Awesome - thanks @anton-seaice! This is a really neat bit of work and it sets up a framework for easily adding other pre- and post-processing steps

anton-seaice:
The only remaining question for me (I think?) is whether to use as is with payu/1.1.3:

```
userscripts:
    setup: /usr/bin/bash /g/data/tm70/as2285/om3-scripts/payu_config/setup.sh
    archive: /usr/bin/bash /g/data/tm70/as2285/om3-scripts/payu_config/archive.sh
```

or with payu/dev:

```
userscripts:
    setup: /usr/bin/bash /g/data/tm70/as2285/om3-scripts/payu_config/setup_scripts/setup_cice_restarts.sh
    archive: /usr/bin/bash /g/data/tm70/as2285/om3-scripts/payu_config/archive_scripts/archive_cice_restarts.sh && /usr/bin/bash /g/data/tm70/as2285/om3-scripts/payu_config/archive_scripts/concat_ice_daily.sh && /usr/bin/bash /g/data/tm70/as2285/om3-scripts/payu_config/archive_scripts/build_intake_ds.py
```

or similar - so we could drop the `setup.sh` and `archive.sh` files ?

dougiesquire:
I think I prefer the former. It means we can update the pre/post-processing without having to update all the configs. So I vote to merge as is.

Issue #16, opened 2024-07-18
`data_stream_xml_generation` scripts produce files that are inconsistent with ACCESS-OM3 configs


dougiesquire:
See https://github.com/COSIMA/access-om3/issues/193

dougiesquire:
Closed by #17

PR #17, opened 2024-07-18
Set zero offset for RYF

ezhilsabareesh8:
This PR addresses #16

### DIFF
--- a/data_stream_xml_generation/generate_xml_datm.py
+++ b/data_stream_xml_generation/generate_xml_datm.py
@@ -84,9 +84,12 @@ for stream_name in stream_info_names:
         "CORE_IAF_JRA55do.LWDN",
         "CORE_IAF_JRA55do.SWDN",
     ]:
-        SubElement(
+        if year_first == year_last:
+            SubElement(stream_info, "offset").text = "0"  #RYF starts at midnight
+        else:    
+            SubElement(
             stream_info, "offset"
-        ).text = "-5400"  # shift back 1.5hr to match RYF
+            ).text = "-5400"  # shift back 1.5hr to match RYF
     else:
         SubElement(stream_info, "offset").text = "0"
 
--- a/data_stream_xml_generation/generate_xml_datm.py
+++ b/data_stream_xml_generation/generate_xml_datm.py
@@ -78,18 +78,15 @@ for stream_name in stream_info_names:
     datafiles = SubElement(stream_info, "datafiles")
     datavars = SubElement(stream_info, "datavars")
 
-    if stream_name in [
+    if stream_name in ([
         "CORE_IAF_JRA55do.PRSN",
         "CORE_IAF_JRA55do.PRRN",
         "CORE_IAF_JRA55do.LWDN",
         "CORE_IAF_JRA55do.SWDN",
-    ]:
-        if year_first == year_last:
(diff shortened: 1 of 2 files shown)


dougiesquire:
Thanks @ezhilsabareesh8. I think `generate_xml_datm.py` also needs a similar change

ezhilsabareesh8:
> Thanks @ezhilsabareesh8. I think `generate_xml_datm.py` also needs a similar change

DATM is fine, both the RYF and IAF dataset starts at midnight. The offset is required only for few variables which starts at 1:30 hrs, which is already implemented in the code.

ezhilsabareesh8:
> Thanks @ezhilsabareesh8. I think `generate_xml_datm.py` also needs a similar change

Fixed it in this commit ae25ac9

data_stream_xml_generation/generate_xml_datm.py
```diff
@@ -84,9 +84,12 @@
...
-        SubElement(
+        if year_first == year_last:
+            SubElement(stream_info, "offset").text = "0"  #RYF starts at midnight
+        else:    
+            SubElement(
             stream_info, "offset"
-        ).text = "-5400"  # shift back 1.5hr to match RYF
+            ).text = "-5400"  # shift back 1.5hr to match RYF
     else:
         SubElement(stream_info, "offset").text = "0"
```
dougiesquire, line 94:
```python
    if (stream_name in [
        "CORE_IAF_JRA55do.PRSN",
        "CORE_IAF_JRA55do.PRRN",
        "CORE_IAF_JRA55do.LWDN",
        "CORE_IAF_JRA55do.SWDN",
    ]) and (year_first != year_last): 
        SubElement(
            stream_info, "offset"
        ).text = "-5400"  # shift back 1.5hr to match RYF
    else:
        SubElement(stream_info, "offset").text = "0"
```

dougiesquire (changes requested):
Thanks @ezhilsabareesh8. One small change for clarity

data_stream_xml_generation/generate_xml_datm.py
ezhilsabareesh8: Changed in this commit [9f28f27](https://github.com/COSIMA/om3-scripts/pull/17/commits/9f28f27ba493547adf0929f38fc08d2bc22f62b9)

data_stream_xml_generation/generate_xml_datm.py
```diff
@@ -78,15 +78,15 @@
...
         "CORE_IAF_JRA55do.PRSN",
         "CORE_IAF_JRA55do.PRRN",
         "CORE_IAF_JRA55do.LWDN",
         "CORE_IAF_JRA55do.SWDN",
-    ]:
+    ]) and (year_first != year_last): 
         SubElement(
             stream_info, "offset"
-        ).text = "-5400"  # shift back 1.5hr to match RYF
+        ).text = "-5400"  # shift back 1.5hr to match RYF
```
dougiesquire, line 89:
```suggestion
        ).text = "-5400"  # shift back 1.5hr to match RYF
```

dougiesquire (changes requested):

dougiesquire (approved):
Thanks @ezhilsabareesh8. LGTM. By the way, you can just accept the suggestions in my review, rather than having to go and make the commits yourself

PR #24, opened 2024-08-21
add make_salt_sfc_restore.py

aekiss:
Adds https://github.com/COSIMA/access-om2/blob/2911891/tools/make_salt_sfc_restore.py

This needs some tweaks to meet the requirements of the README, particularly adding provenance metadata. @ezhilsabareesh8 would you have time to do that before we merge this PR?

### DIFF
--- a/salt_sfc_restore_generation/make_salt_sfc_restore.py
+++ b/salt_sfc_restore_generation/make_salt_sfc_restore.py
@@ -1,0 +1,176 @@
+#!/usr/bin/env python
+
+import numpy as np
+import sys
+import numba
+import os
+import argparse
+import netCDF4 as nc
+import shutil
+from scipy.ndimage.filters import uniform_filter
+from make_remap_weights import create_weights
+
+from esmgrids.mom_grid import MomGrid  # noqa
+from esmgrids.woa_grid import WoaGrid  # noqa
+
+"""
+Create MOM salt_sfc_restore.nc file from WOA
+"""
+
+def calc_regridding_err(weights, src, dest):
+    """
+    Calculate the regridding error.
+    """
+
+    EARTH_RADIUS = 6370997.0
+
+    with nc.Dataset(weights) as f:
+        try:
+            area_a = f.variables['area_a'][:]
+        except KeyError as e:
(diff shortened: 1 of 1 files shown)


aekiss:
related: https://github.com/COSIMA/access-om3/issues/207

aekiss:
From the metadata it looks like this was also needed. We should include it in the script so it's a one-step process.
```
ncatted -O -a units,time,m,char,months since 0001-01-01 00:00:00 salt_sfc_restore.nc
ncrename -O -v so,salt mean_salinity_of_0_and_10m.nc
```

```
ncdump -h /g/data/ik11/inputs/access-om2/input_20201102/mom_01deg/salt_sfc_restore.nc
netcdf salt_sfc_restore {
...
// global attributes:
		:comment = "This salinity distribution is based on WOA13v2 where the land values are filled in with a conjugate gradient method. It is a simple average of land-filled salinity data at the depth of 0 and 10 meters." ;
		:history = "Fri Jun 28 12:31:42 2019: ncatted -O -a units,time,m,char,months since 0001-01-01 00:00:00 salt_sfc_restore.nc\nFri Dec  1 14:47:24 2017: ncrename -O -v so,salt mean_salinity_of_0_and_10m.nc" ;
		:NCO = "netCDF Operators version 4.7.7 (Homepage = http://nco.sf.net, Code = http://github.com/nco/nco)" ;
}
```

aekiss:
Oops, converting to draft. I'm not sure this is a good way to go, since this script also needs a lot of other code from https://github.com/COSIMA/access-om2/blob/master/tools. Do we want all that here as well?

ezhilsabareesh8:
> @ezhilsabareesh8 would you have time to do that before we merge this PR?

@aekiss Yes I can do that, also there are some dependencies required for `make_salt_sfc_restore.py`, for example [make_remap_weights.py](https://github.com/COSIMA/access-om2/blob/master/tools/make_remap_weights.py), which also needs to be included if we move `make_salt_sfc_restore.py` to `om3-scripts`.

anton-seaice:
We decided to use https://github.com/ACCESS-NRI/initial_conditions_access-om3/

PR #25, opened 2024-08-23
allow for default initial conditions

anton-seaice:
Allow for default initial conditions per https://github.com/COSIMA/MOM6-CICE6/pull/103#discussion_r1728057112


dougiesquire (approved):
Thanks @anton-seaice. LGTM

Issue #40, opened 2024-11-01
elementArea field in mesh fieles has wrong units


anton-seaice:
It looks like we should be using units of radians^2 for _elementArea_ for in ESMF unstructed mesh files.

e.g.
```
     double elementArea(elementCount) ;
            elementArea:units = "radians^2" ;
            elementArea:long_name = "area weights" ;
```

See:

https://earthsystemmodeling.org/docs/release/ESMF_8_0_1/ESMF_refdoc/node3.html#SECTION03028200000000000000

Saying that, we are not currently using this field. e.g. 

https://github.com/CICE-Consortium/CICE/blob/3ebdef778bded4b10d628a179ea746d091e21150/cicecore/drivers/nuopc/cmeps/ice_comp_nuopc.F90#L732-L733

would need [another argument](https://github.com/esmf-org/esmf/blob/bd7a15d782b6f2ba044f197a49b87278224feaf5/src/Infrastructure/Mesh/interface/ESMF_Mesh.F90#L1860C58-L1860C69) ( `addUserArea = .true.` ) to actually load this. (See https://github.com/COSIMA/access-om3/issues/240)

It also looks like the cells at the north poles on the grid, should be using numElementConn = 3 , not 4. 

i.e.

_elementConn can be either a 2D array or a 1D array. If it is a 2D array, the second dimension of the elementConn variable has to be the size of the largest number of nodes in any element (i.e. maxNodePElement), the actual number of nodes in an element is given by the numElementConn variable._

<img width="1555" alt="Screenshot 2024-11-01 at 10 48 54 AM" src="https://github.com/user-attachments/assets/e3a35b93-c492-4151-9387-b3b77c32626e">

Presumable the last element in the _elementConn_ field is set the the `_FillValue` when there are only three corners to an element

```
 int elementConn(elementCount, maxNodePElement) ;
            elementConn:long_name = "Node Indices that define the element /
                                     connectivity";
            elementConn:_FillValue = -1 ;
            elementConn:start_index = 1 ;
```

anton-seaice:
The area field did get fixed in #41 

There is also 

> It also looks like the cells at the north poles on the grid, should be using numElementConn = 3 , not 4.

Not sure if it affects anything or not

PR #41, opened 2024-11-01
Use radians^2 for cell areas

anton-seaice:
Contributes to #40 

Use radians^2 for area, and fix metadata


ezhilsabareesh8 (approved):
Thanks @anton-seaice. Looks good to me

Issue #44, opened 2025-01-16
add rof weights generation script


anton-seaice:
As described [here](https://github.com/COSIMA/access-om3/issues/231#issuecomment-2579167824) add a script to make remapping weights which map runoff from land cells to ocean cells and conserve the total volume of runoff. This will use the esmf mesh format used by the OM3 mediator.

aekiss:
FYI here's the runoff misdirection problem we found previously with a kdtree method when the nearest ocean point is on the other side of a long skinny land barrier https://github.com/COSIMA/access-om2/issues/130

anton-seaice:
Thanks!

It's possible a brute-force approach rather than a KD-tree would be fine - we only have to make the weights once. The brute-force approach would allow for a more accurante surface of a sphere distances in metres rather than euclidean (in degrees). It also might automate some cleverer handling of the nearest cell to preference cells connected through the ocean, rather than over land (although it would take some thinking to know how to include the JRA & deseas landmasks etc).

anton-seaice:
Closed in #45

PR #45, opened 2025-01-16
remapping weights script

anton-seaice:
Contributes to #44. The script makes remapping weights which map from all land cells into the nearest ocean cell and conserves the total volume. There is no spreading of runoff, or effort to put the runoff at the mouth of a bay/river when it is further away than the closest ocean.

Note:

There is not enough memory on the gadi login node to run this, its simplest to run in a terminal through are.nci.org.au

I get this warning, but it works ok:

```
WARNING: There was an error initializing an OpenFabrics device.

  Local host:   gadi-cpu-clx-1209
  Local device: mlx5_0
```


mesh_generation/generate_rof_weights.py
```diff
@@ -0,0 +1,172 @@
+# Copyright 2025 ACCESS-NRI and contributors. See the top-level COPYRIGHT file for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# =========================================================================================
+# Generate an remapping weights between two ESMF mesh files for remapping a runoff field
+# an unmasked mesh to a masked mesh without loosing any water volume. Each field on the
```
aekiss, line 6:
```suggestion
# an unmasked mesh to a masked mesh without losing any water volume. Each field on the
```

mesh_generation/generate_rof_weights.py
```diff
@@ -0,0 +1,172 @@
...
+path_root = Path(__file__).parents[1]
+sys.path.append(str(path_root))
+from scripts_common import get_provenance_metadata, md5sum
+
+TEMP_WEIGHTS_F = "temp_weights.nc"
+COMP_ENCODING = {"complevel": 1, "compression": "zlib"}  # compression settings to use
+
+
+def drof_remapping_weights(mesh_filename, weights_filename, global_attrs=None):
+    # We need to generate remapping weights for use in the mediator, such that the overall volume of runoff is conserved and no run-off is mapped onto land cells. Inside the mediator, the grid doesn't change as we run the mediator with the ocean grid (the DROF component does the remapping from JRA grid to mediator grid). There we use the same _mesh_file for the input and output mesh, however this same routine would work for differeing input and output meshes
```
aekiss, line 50:
```suggestion
    # We need to generate remapping weights for use in the mediator, such that the overall volume of runoff is conserved and no runoff is mapped onto land cells. Inside the mediator, the grid doesn't change as we run the mediator with the ocean grid (the DROF component does the remapping from JRA grid to mediator grid). There we use the same _mesh_file for the input and output mesh, however this same routine would work for differing input and output meshes
```

mesh_generation/generate_rof_weights.py
```diff
@@ -0,0 +1,172 @@
...
+    )
+
+    med_in_fld = esmpy.Field(model_mesh, meshloc=esmpy.MeshLoc.ELEMENT)
+
+    med_out_fld = esmpy.Field(model_mesh, meshloc=esmpy.MeshLoc.ELEMENT)
+
+    try:
+        os.remove(TEMP_WEIGHTS_F)  # rm old temp file
+    except:
+        None
```
aekiss, line 64:
```suggestion
    except OSError:
        pass
```
https://stackoverflow.com/questions/10840533/most-pythonic-way-to-delete-a-file-which-may-not-exist

mesh_generation/generate_rof_weights.py
```diff
@@ -0,0 +1,172 @@
...
+
+    # Find index for all ocean cells
+    mask_i = mod_mesh_ds.elementCount.where(mod_mesh_ds.elementMask, drop=True).astype(
+        "int"
+    )
+
+    # Make a KDTree from the ocean cells
+    mask_tree = KDTree(mod_mesh_ds.centerCoords.isel(elementCount=mask_i))
+
+    # Using the KDTree, look up the nearest ocean cell to every destination grid cell in our weights file. Note our weights are indexed from 1 (i.e. Fortran style) but xarray starts from 0 (i.e. python style), so subract one from our destination grid cell indices.
```
aekiss, line 95: Here, does "nearest" mean "smallest difference in grid index"? This may not correspond to smallest distance on the sphere, especially at the join between the eastern and western edges, and in the tripolar region.

mesh_generation/generate_rof_weights.py
```diff
@@ -0,0 +1,172 @@
...
+    area = copy(med_out_fld.data)
+    old_area = area[weights_ds.row - 1]
+    new_area = area[new_row - 1]
+
+    weights_ds["row"] = xr.DataArray(data=new_row, dims="n_s")
+
+    weights_ds["S"] = weights_ds.S * old_area / new_area
+
+    # add global attributes
+    weights_ds.attrs = {
```
aekiss, line 115: I thought we had some shared code to do this provenance stuff?

mesh_generation/generate_rof_weights.py
```diff
@@ -0,0 +1,172 @@
+# Copyright 2025 ACCESS-NRI and contributors. See the top-level COPYRIGHT file for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# =========================================================================================
+# Generate an remapping weights between two ESMF mesh files for remapping a runoff field
+# an unmasked mesh to a masked mesh without loosing any water volume. Each field on the
+# unmasked mesh is mapped to the nearest ocean cell in the resulting weights.
+#
+# To run:
+#   python generate_rof_weights.py --mesh_filename=<input_file> --weights_filename=<output_file>
```
aekiss, line 10: could drop both instances of `_filename` for brevity?

mesh_generation/generate_rof_weights.py
anton-seaice: `centerCoords` is the geographic coordinate of the center of the cell. This function returns the index of the cell with the smallest difference in the geographic coordinate (by euclidean distance, not surface of a sphere). It won't wrap from east to western edge but should be ok for everywhere else.

mesh_generation/generate_rof_weights.py
anton-seaice: We used _filename in `generate_mesh.py`, so I went with that.

It also make writing the code easier, as there are mesh objects made from mesh files in the script

mesh_generation/generate_rof_weights.py
```diff
@@ -0,0 +1,172 @@
...
+
+    weights_ds["row"] = xr.DataArray(data=new_row, dims="n_s")
+
+    weights_ds["S"] = weights_ds.S * old_area / new_area
+
+    # add global attributes
+    weights_ds.attrs = {
+        "gridType": "unstructured mesh",
+        "timeGenerated": f"{{str(datetime.now())[0:16]}}",
+        "created_by": f"{os.environ.get('USER')}",
```
anton-seaice, line 118:
```suggestion
```

Duplicated from global_attrs

aekiss (approved):
Thanks @anton-seaice, looks good. My only significant concern is whether this is nearest-neighbour on the sphere or on the grid.

mesh_generation/generate_rof_weights.py
aekiss: ok, might need to take a look at whether anything fishy happens at the join - I hope there are no major rivers near 80E in the Indian or Arctic ...

anton-seaice:
> Thanks @anton-seaice, looks good. My only significant concern is whether this is nearest-neighbour on the sphere or on the grid.

Its nearest-neighbour by euclidean distance between the center coordinates (in degrees) of the grid. This means that for the non-square cells, the results would be a bit weird (e.g. a nominal 0.25deg cell might be 6km in x and 11km in y in the Arctic, but be around 0.25 degrees in both directions).

anton-seaice:
@aekiss 


It looks like using a _Ball Tree_ instead of a _KDTree_ allows for 'haversine' distances

See https://scikit-learn.org/stable/modules/neighbors.html#ball-tree

and 
https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.haversine_distances.html#sklearn.metrics.pairwise.haversine_distances

The haversine distances are appropriate for surface of the sphere, and should wrap east to west around the globe. Its a fairly simple change by the looks of it.

aekiss:
Cool, I don't think the difference between nearest-by-degrees and by-distance is likely to make much difference, but we may as well try us eball tree if it's easy.

aekiss:
Oh, I just re-read - if it also maps at the east-west join, then we should definitely use it. Will it also handle mapping across the tripole seam?

anton-seaice:
I made the update - the new scheme is slower to make weights but otherwise very similar (and still conserves runoff in the model)

mesh_generation/generate_rof_weights.py
```diff
@@ -0,0 +1,177 @@
...
+    # Find index for all ocean cells
+    mask_i = mod_mesh_ds.elementCount.where(mod_mesh_ds.elementMask, drop=True).astype(
+        "int"
+    )
+
+    center_coords_rad = deg2rad(mod_mesh_ds.centerCoords)
+
+    # Make a BallTree from the ocean cells
+    mask_tree = BallTree(
+        center_coords_rad.isel(elementCount=mask_i), metric="haversine"
```
dougiesquire, line 97:
```suggestion
        center_coords_rad.isel(elementCount=mask_i, coordDim=[1,0]), metric="haversine"
```

mesh_generation/generate_rof_weights.py
```diff
@@ -0,0 +1,177 @@
...
+    )
+
+    # Using the Tree, look up the nearest ocean cell to every destination grid cell in our weights file. Note our weights are indexed from 1 (i.e. Fortran style) but xarray starts from 0 (i.e. python style), so subract one from our destination grid cell indices.
+
+    ii = mask_tree.query(
+        center_coords_rad.isel(elementCount=(weights_ds.row - 1)), return_distance=False
```
dougiesquire, line 103:
```suggestion
        center_coords_rad.isel(elementCount=(weights_ds.row - 1), coordDim=[1,0]), 
        return_distance=False
```

dougiesquire (changes requested):
Thanks @anton-seaice!

I created a plot to convince myself that the BallTree query is working as expected and it revealed a small bug.

In the following, crosses shows every 50th cell centre. Blue crosses are land cells, yellow crosses are ocean cells. Red lines show which land cells are mapped to which ocean cells. Clearly there are land cells in the Antarctic being mapped to ocean cells in the Arctic.

Screenshot 2025-01-22 at 7 56 29 pm

As we worked out in our chat, this is because the coordinates need to be provided to the BallTree routines in the order `(lat, lon)`, not `(lon, lat)`. Switching this, as suggested in my review, fixes the issue.

Screenshot 2025-01-22 at 8 03 58 pm

mesh_generation/generate_rof_weights.py
dougiesquire: Using the haversine distance (with a small bug fix) successfully maps across the join - see https://github.com/COSIMA/om3-scripts/pull/45#pullrequestreview-2566520736

anton-seaice:
Thanks @dougiesquire - well spotted. I made the fix :)

image

dougiesquire (approved):
LGTM - nice work!

aekiss:
Great work both, LGTM!

PR #49, opened 2025-02-27
add vertical grid generation script

ezhilsabareesh8:
This PR addresses issue [249](https://github.com/COSIMA/access-om3/issues/249).

Added a vertical grid generation script based on Kial's KDS75 vertical grid generation script.

### DIFF
--- a/grid_generation/generate_vertical_grid.py
+++ b/grid_generation/generate_vertical_grid.py
@@ -70,6 +70,7 @@ min_dz = args.min_dz
 depfac = args.depfac
 output_filename = args.output
 
+# Define a small constant to initialize the iteration and prevent numerical issues
 epsilon = 0.001
 
 
--- a/grid_generation/generate_vertical_grid.py
+++ b/grid_generation/generate_vertical_grid.py
@@ -50,8 +50,8 @@ def f_all(kk, H, depfac, dzd):
 
 def generate_vertical_grid(H, dzd, min_dz, depfac, output_filename):
 
-# Initialize the first two entries of the vertical grid with 0 and epsilon 
-# for both depth (z) and grid spacing (dz).
+    # Initialize the first two entries of the vertical grid with 0 and epsilon 
+    # for both depth (z) and grid spacing (dz).
     delta_z = [0, epsilon]
     prop_z = [0, epsilon]
 
--- a/grid_generation/generate_vertical_grid.py
+++ b/grid_generation/generate_vertical_grid.py
@@ -1,10 +1,11 @@
-# Copyright 2023 ACCESS-NRI and contributors. See the top-level COPYRIGHT file for details.
+# Copyright 2025 ACCESS-NRI and contributors. See the top-level COPYRIGHT file for details.
 # SPDX-License-Identifier: Apache-2.0
 
 # =========================================================================================
-# Generate a vertical grid for ocean models using a hyperbolic tangent function.
+# Generate a vertical grid for MOM using a hyperbolic tangent function.
 #
-# This script creates a regular vertical grid and writes it to a NetCDF file.
+# This script creates a vertical grid and writes it to a NetCDF file.
+# This script relates to the article "Stewart, K.D., Hogg, A.M., Griffies, S.M., Heerdegen, A.P., Ward, M.L., Spence, P. and England, M.H., 2017. Vertical resolution of baroclinic modes in global ocean models. Ocean Modelling, 113, pp.50-65."
 #
 # Usage:
 # Run the script with command-line arguments to specify:
@@ -113,7 +114,7 @@ eddyfile = nc.Dataset(output_filename, "w", format="NETCDF4")
 eddyfile.createDimension("nzv", len(real_prop_z))
 v_grid = eddyfile.createVariable("v_grid", "f8", ("nzv",))
 v_grid.units = "meters"
-v_grid.standard_name = "vertical_grid"
+v_grid.standard_name = "cell_thickness"
 v_grid.long_name = "vertical grid depth at top and bottom of each cell"
 eddyfile.variables["v_grid"][:] = real_prop_z
 eddyfile.setncatts({"history": get_provenance_metadata(this_file, runcmd)})
--- a/grid_generation/generate_vertical_grid.py
+++ b/grid_generation/generate_vertical_grid.py
@@ -6,8 +6,8 @@
 #
 # This script creates a vertical grid and writes it to a NetCDF file.
 # This script relates to the article "Stewart, K.D., Hogg, A.M., Griffies, S.M., Heerdegen, A.P.,
- Ward, M.L., Spence, P. and England, M.H., 2017. Vertical resolution of baroclinic modes in 
- global ocean models. Ocean Modelling, 113, pp.50-65."
+# Ward, M.L., Spence, P. and England, M.H., 2017. Vertical resolution of baroclinic modes in
+# global ocean models. Ocean Modelling, 113, pp.50-65. http://dx.doi.org/10.1016/j.ocemod.2017.03.012"
 #
 # Usage:
 # Run the script with command-line arguments to specify:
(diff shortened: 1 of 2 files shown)


vertical_grid_generation/generate_vertical_grid.py
```diff
@@ -0,0 +1,124 @@
...
+    f"--dzd={dzd} "
+    f"--min_dz={min_dz} "
+    f"--output={output_filename} "
+)
+
+eddyfile = nc.Dataset(output_filename, "w", format="NETCDF4")
+eddyfile.createDimension("nzv", len(real_prop_z))
+v_grid = eddyfile.createVariable("v_grid", "f8", ("nzv",))
+v_grid.units = "meters"
+v_grid.standard_name = "vertical_grid"
```
anton-seaice, line 116: standard_name should use the cf_convention for standard_name, which might be "cell_thickness" ?

https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html

anton-seaice (changes requested):
I haven't run this yet - i left some general comments. Mostly I think the structure could be tidied up and some comments added. 

Just confirming - this script is expected to make the grid as described in https://github.com/COSIMA/access-om3/issues/249#issuecomment-2549920512 ?

I think it's worth asking @aekiss for a quick look - especially that the vertical grid is appropriate.

ezhilsabareesh8:
> Just confirming - this script is expected to make the grid as described in [COSIMA/access-om3#249 (comment)](https://github.com/COSIMA/access-om3/issues/249#issuecomment-2549920512) ?

Yes it is the same, and all the comments are addressed.

anton-seaice (commented):
Thanks @ezhilsabareesh8 

I think this could be better structured and needs more comments.

A minimal way could be to follow the _generatre_rof_weights.py_ script. - Which has a main() function and a function to do the calculating.

Then for each function, a comment or docstring describing would it does is needed.

In this case - because there is some specific algorithm being implemented - a link to a paper of where the algorithm comes from could be good. Or otherwise, comments to describe the processes undertaken.

grid_generation/generate_vertical_grid.py
```diff
@@ -0,0 +1,138 @@
...
+      f"Grid spacing ranges from {real_delta_z[0]:.2f} m at the surface to {real_delta_z[-1]:.2f} m at depth. "
+      f"Output written to: {output_filename}")
+
+def write_netcdf_file(output_filename, real_prop_z, this_file, runcmd):
+    """Function to write vertical grid data to a NetCDF file."""
+    eddyfile = nc.Dataset(output_filename, "w", format="NETCDF4")
+    eddyfile.createDimension("nzv", len(real_prop_z))
+    zeta = eddyfile.createVariable("zeta", "f8", ("nzv",))
+    zeta.units = "meters"
+    zeta.standard_name = "cell_thickness"
```
anton-seaice, line 103: Sorry - I might have misunderstood what was in this file. I thought it had cell thicknesses but it actually has depths. 

A better `standard_name` might be `depth_below_geoid` ?

There is this [note](https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html) for the field: 

`The geoid is a surface of constant geopotential with which mean sea level would coincide if the ocean were at rest. (The volume enclosed between the geoid and the sea floor equals the mean volume of water in the ocean). In an ocean GCM the geoid is the surface of zero depth, or the rigid lid if the model uses that approximation. To specify which geoid or geopotential datum is being used as a reference level, a grid_mapping variable should be attached to the data variable as described in Chapter 5.6 of the CF Convention. "Depth_below_X" means the vertical distance below the named surface X.`

Do you know @aekiss - which geoid/ geopotential datum we use for 0 sea level ?

anton-seaice (commented):
This looks ok to me and runs ok ! It would be good to get @aekiss to look

I just noticed that the old v_grid has been changed to do that the numbers are single precision but still saved at double precision. Is this still important for MOM6?

```
NOTE = "Modified by Russ Fiedler. Suitable for double and single precision arithmetic" ;
                :history = "Thu Sep 19 17:04:43 2024: ncks -O -d nzv,,,2 /g/data/ik11/inputs/access-om2/input_20201102/mom_01deg/ocean_vgrid.nc ocean_vgrid_75.nc\nMon Jul 24 14:14:08 2017: ncatted -a NOTE,global,c,c,Modified by Russ Fiedler. Suitable for double and single precision arithmetic ocean_vgrid.nc" ;
```

grid_generation/generate_vertical_grid.py
```diff
@@ -0,0 +1,138 @@
+# Copyright 2025 ACCESS-NRI and contributors. See the top-level COPYRIGHT file for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# =========================================================================================
+# Generate a vertical grid for MOM using a hyperbolic tangent function.
+#
+# This script creates a vertical grid and writes it to a NetCDF file.
+# This script relates to the article "Stewart, K.D., Hogg, A.M., Griffies, S.M., Heerdegen, A.P., Ward, M.L., Spence, P. and England, M.H., 2017. Vertical resolution of baroclinic modes in global ocean models. Ocean Modelling, 113, pp.50-65."
```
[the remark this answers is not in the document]
aekiss: also include doi:
http://dx.doi.org/10.1016/j.ocemod.2017.03.012

grid_generation/generate_vertical_grid.py
aekiss: no particular datum is used (AFAIK we don't even take oblateness of the Earth into account)

aekiss:
> I just noticed that the old v_grid has been changed to do that the numbers are single precision but still saved at double precision. Is this still important for MOM6?

I think it would be good to have double precision values that are exactly representable in single precision. It might bite us somehow if not.

Russ had a separate Fortran tool to do this

aekiss:
looks ok to me on a quick skim

ezhilsabareesh8:
Thanks @anton-seaice and @aekiss, I have incorporated all the suggestions.

aekiss (approved):
LGTM, thanks @ezhilsabareesh8

grid_generation/generate_vertical_grid.py
anton-seaice: Ah yes - the modelled earth is round ...

And these depths are totally independent of whether its an earth or a regional model or a different planet anyway, so I guess just set this standard_name to `depth` ?

anton-seaice:
> > I just noticed that the old v_grid has been changed to do that the numbers are single precision but still saved at double precision. Is this still important for MOM6?
> 
> I think it would be good to have double precision values that are exactly representable in single precision. It might bite us somehow if not.
> 
> Russ had a separate Fortran tool to do this, so we could just continue to do that unless you want to incorporate that feature in this python code.

Can you implement this please @ezhilsabareesh8 ?

It looks like [all thats needed](https://github.com/COSIMA/topogtools/blob/6b4983127aa18dfdf1b62b2e18b581d82d4a64d4/float_vgrid.f90#L18-L19) is to save it as a type single and then back as a type double, but please check it works

grid_generation/generate_vertical_grid.py
```diff
@@ -0,0 +1,168 @@
...
+    # Convert to float32 (single precision) to ensure values are exactly representable in single precision,
+    # then convert back to float64 (double precision) for storage in NetCDF.
+    real_prop_z_float32 = real_prop_z.astype(np.float32)
+    real_prop_z_float64 = real_prop_z_float32.astype(np.float64)
+
+    eddyfile = nc.Dataset(output_filename, "w", format="NETCDF4")
+    eddyfile.createDimension("nzv", len(real_prop_z))
+    zeta = eddyfile.createVariable("zeta", "f8", ("nzv",))
+    zeta.units = "meters"
+    zeta.standard_name = "depth_below_geoid"
```
anton-seaice, line 125:
```suggestion
    zeta.standard_name = "depth"
```

anton-seaice (approved):
Thanks @ezhilsabareesh8 - looks good !

PR #51, opened 2025-03-12
use xp65 instead of hh5 for new versions

### LINKED ISSUE #50: Don't use `hh5` conda environment in userscripts
We're currently using the `hh5` environment to auto-generate intake-esm datastores in [`payu_config/archive_scripts/build_intake_ds.sh`](https://github.com/COSIMA/om3-scripts/blob/main/payu_config/archive_scripts/build_intake_ds.sh):

https://github.com/COSIMA/om3-scripts/blob/6303549049cc0ad725fc4d2938b7aa6408992dc0/payu_config/archive_scripts/build_intake_ds.sh#L8-L9 

This environment is no longer maintained and is using an out-of-date version of `access-nri-intake`. This is [causing issues for users](https://forum.access-hive.org.au/t/access-rom3-setup-instructions/4107/80?u=dougiesquire).

The long term aim is to have the code to generate datastores [included in Payu itself](https://github.com/payu-org/payu/issues/521), but we should move to using the `xp65` environment instead of `hh5` as a stop-gap.

anton-seaice:
Closes #50

### DIFF
--- a/payu_config/archive.sh
+++ b/payu_config/archive.sh
@@ -2,4 +2,4 @@ source $(dirname "$0")/archive_scripts/archive_cice_restarts.sh
 source $(dirname "$0")/archive_scripts/concat_ice_daily.sh
 source $(dirname "$0")/archive_scripts/standardise_mom6_filenames.sh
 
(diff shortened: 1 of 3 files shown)


dougiesquire (approved):
Looks good and tested this successfully. Thanks @anton-seaice

Issue #61, opened 2025-06-06
Coupler restart file is overwritten during update


minghangli-uni:
When updating coupler restart files, the current `remask_cpl_restart.py` overwrites the existing restart file by default. This may not be safe, as users might want to preserve the original file for comparison or backup purposes.

To improve usability and prevent accidental data loss, it would be better to implement an option to either retain the original file or write to a new output path by default.

PR #62, opened 2025-06-06
Add --output_file to write fixed restart to a separate file

minghangli-uni:
Closes https://github.com/ACCESS-NRI/om3-scripts/issues/61

This PR introduces a required `--output_file` argument to `remask_cpl_restart.py`, enabling users to write the processed coupler restart file to a separate location instead of modifying the original input in-place.


ezhilsabareesh8 (approved):
Thanks @minghangli-uni, LGTM

Issue #63, opened 2025-06-10
Incorrect filename format after combining MOM6 output with non-default I/O layout


minghangli-uni:
When using a MOM I/O layout other than the default 1,1, the model produces multiple output files such as

```
access-om3.mom6.3d.vh.1mon.mean._1900.nc.0001
access-om3.mom6.3d.vh.1mon.mean._1900.nc.0002
access-om3.mom6.3d.vh.1mon.mean._1900.nc.0003
access-om3.mom6.3d.vh.1mon.mean._1900.nc.0004
access-om3.mom6.3d.vh.1mon.mean._1900.nc.0005
```

After combining these diags using https://github.com/coecms/mppnccombine-fast, the output will be like,

```
access-om3.mom6.3d.vh.1mon.mean._1900.nc
```

which still contains a "_" before year.

PR #64, opened 2025-06-11
Fix mom6 filename format with non-default I/O layout

minghangli-uni:
Closes https://github.com/ACCESS-NRI/om3-scripts/issues/63

Update below two scripts,

- [x] standardise_mom6_filenames.sh
- [x] test_mom6_filenames.py


minghangli-uni:
@dougiesquire could you review this when you are available?

dougiesquire (approved):
Thanks @minghangli-uni. LGTM. Ideally we would fix this in FMS but that's a problem for another day.

PR #67, opened 2025-06-19
Add make_ryf code

anton-seaice:
In the RYF files we use, the timestamps are defined at the start of each time interval. However in the source JRA55do data, the timestamps are at the start of the interval for states/tracers and are in the middle of each time interval for fluxes. For consistency, this PR updates the make RYF code to be consistent with the original data.

In addition, use the latest JRA55do, although there are no changes to the data between versions 1.5/1.6. 1.6 is latest and final release, and is the same as 1.5 and the 1.5 extensions, just published in one dataset.

Code and commits merged from https://github.com/COSIMA/make_ryf/ but included in this repo to make access to the "scripts_common" functions easier. Thankyou to @aekiss and @aidanheerdegen for the work here!

This address https://github.com/ACCESS-NRI/access-om3-configs/issues/387, https://github.com/ACCESS-NRI/access-om3-configs/issues/386 and https://github.com/ACCESS-NRI/access-om3-configs/issues/437. 

(In my view - scope of review is the final commit, where changes were made from COSIMA code)

The generated RYF data has not changes in data compared to the 1.4 version, see https://github.com/anton-seaice/sandbox/blob/3e11e7f8e46437ca94beb5cd03107e5aa824a8f8/ryf_comparo_1.6.ipynb


anton-seaice:
@dougiesquire - this is ready for review. If the gist of this looks ok, ill run some tests with the model

anton-seaice:
We agreed yesterday at TWG to use linear interpolation, so this is ready @dougiesquire 

example stream files in https://github.com/ACCESS-NRI/access-om3-configs/pull/615 and https://github.com/ACCESS-NRI/access-om3-configs/pull/616

dougiesquire (dismissed):
Thanks @anton-seaice!

dougiesquire (approved):

PR #69, opened 2025-06-25
Concat daily cice output in python

anton-seaice:
I've drafted a possible python implementation for concat ice daily to addres https://github.com/ACCESS-NRI/access-om3-configs/issues/501

A couple of issues:
- it doesn't feel any faster than the nco method
- dask is not in payu/condaenv 

Is this the kind of approach you were thinking @dougiesquire ? I will test some more with larger jobs and more reaslistic test data, but not quite sure of the right direction from here


anton-seaice:
@dougiesquire - I tried changing chunk cache and usind dask delayed objects etc, but none of it is faster than using the `ncrcat` approach or completes with hdf5 errros. See notebook in this PR.

I'm out of ideas ? I think we might just need to do this in a seperate qsub job ?

dougiesquire (commented):
Thanks @anton-seaice. Sorry for the delay. The approach looks sensible to me. Did you say that it was the setting of `PYTHONNOUSERSITE=True` that fix the performance issue relative to nco for you?

Also, just noting that dask's `LocalCluster` can't utilise more than one node. You say this above, so I think you know it, but just noting because the qsub script currently requests two nodes.

anton-seaice:
> Thanks @anton-seaice. Sorry for the delay. The approach looks sensible to me. Did you say that it was the setting of `PYTHONNOUSERSITE=True` that fix the performance issue relative to nco for you?

This fixed the hdf5 errors, so the code ran to completion.

> Also, just noting that dask's `LocalCluster` can't utilise more than one node. You say this above, so I think you know it, but just noting because the qsub script currently requests two nodes.

Do you have a preference for archive userscript vs postscript

dougiesquire:
Both aren't ideal huh.

> 2. make this a post-script. This has the disadvantage of preventing payu running the sync step automatically, and may mean that the post processing gets stuck in the queue and takes a long time to start.

Am I right in understanding that the sync directory will just not include the latest run?

I'm inclined towards the first option I think? 3mins of idle 1560 cores wastes 156 SU?

anton-seaice:
I was thinking of adding the postscript anyway for intake datastore generation - https://github.com/ACCESS-NRI/om3-scripts/issues/70 ?

dougiesquire:
Oh yeah true. Well in that case it's an easy decision

anton-seaice:
intake datastore generation is still parallelised right ?

anton-seaice:
Just checking it makes sense to run it in a 104 core job for a short time (possibly on the express queue) rather than a small job for a longer time

charles-turner-1:
I haven't changed anything about the implementation of generating the datastores, but I think the parallelisation is written such that you might not get any benefit if you build a single esm datastore?

dougiesquire:
If nothing has changed, it should parallelise the file-trawling

charles-turner-1:
Sweet. Max mentioned [this repo](https://github.com/klindsay28/esm_catalog_utils) to me at Scipy - apparently there are some performance improvements in it, relative to ecgtools. I must have gotten the wrong end of the stick about where the improvements come from though.

anton-seaice:
This is ready for review @dougiesquire - the main issue I have is that if dask compute fails, then it will continue and delete the daily files, even though it hasn't produce output successfully

anton-seaice:
I tweaked up the error handling and fixed up the tests.

This now compares the monthly output with the daily input before deleting any data.

dougiesquire (approved):
Thanks @anton-seaice - looks great. Apologies, when I originally suggested checking that the output matches the input, I was just imagining checking that the data from `open_mfdataset(daily_files)` matches `open_mfdataset(monthly_files)`. But I see it has to be more complicated due to potentially incomplete months.

Have you done any testing on timing relative to nco now that the input/output check is in there?

anton-seaice:
The main trick was remembering that `NAN==NAN` evaluates as false. But then using `xr.testing.assert_equal` didn't parallelise very well and was slow, so i split it by month and variable.

Timing is less than 10 minutes currently for two years of 25km data, I think nco was taking about 40 mins.

Maybe I could simplify a little by just overwriting if there are monthly files already, and not ignoring incomplete months, but here we are !

Issue #73, opened 2025-07-28
Improve performance for daily cice concatenation


minghangli-uni:
Running `concat_ice_daily.py` takes around 20 minutes for a 3-year output because each daily netcdf file is loaded as a separate dask chunk. Re-chunking along the time dimension (31 days per chunk) before downstream operations can reduce task-graph size and eliminates repeated I/O. The time using can be reduced to 7 minutes for the results.

PR #74, opened 2025-07-28
Rechunk and persisting daily ice data

minghangli-uni:
closes https://github.com/ACCESS-NRI/om3-scripts/issues/73

This improves the performance of the `concat_ice_daily.py` by rechunking the daily dataset along the time dimension into 31 day blocks, and persisting the rechunked dataset to memory before further processing.

For example, processing 3 years of daily cice output now takes ~7 minutes, down from 20 minutes.

### DIFF
--- a/payu_config/postscript/concat_ice_daily.py
+++ b/payu_config/postscript/concat_ice_daily.py
@@ -127,7 +127,7 @@ class Concat_Ice_Daily:
             raise Exception(f"No daily output files found in {directory}")
 
         self.client = start_client(assume_gadi)
-        daily_ds = lazy_open(self.daily_f)
+        daily_ds = lazy_open(self.daily_f).chunk({"time": 31}).persist()
 
         # del incorrect metadata
         del daily_ds.attrs["comment2"]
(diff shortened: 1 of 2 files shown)


anton-seaice:
Oh that's crazy!

The problem with hardcoding chunksize to 31 is that it doesn't align with months anymore.

What happens if you put the persist at line ~181? Or sometime after the chunksize is set to the length of the month ?

minghangli-uni:
> The problem with hardcoding chunksize to 31 is that it doesn't align with months anymore.

I dont think that matters. It is just for memory and compute efficiency. I think this is a @dougiesquire's question here. 

> What happens if you put the persist at line ~181? Or sometime after the chunksize is set to the length of the month ?

Do you mean inside the monthly_pairs loop? I think that's gonna waste effort since the data is not used anymore after loading.

anton-seaice (approved):
Thanks @minghangli-uni 

The chunking is changed again at line 182 to match the length of the month, however setting the chunking to length 31 seems to help when doing the lazy loads etc before 182.

persist is a good pickup! I can only assume it's inherited by the datasets in `self.month_ds` - which are the ones actually loaded into memory!

PR #75, opened 2025-07-31
accepts topog instead of ocean_mask

minghangli-uni:
closes https://github.com/ACCESS-NRI/access-om3-configs/issues/311

### DIFF
--- a/mesh_generation/generate_mesh.py
+++ b/mesh_generation/generate_mesh.py
@@ -510,7 +510,10 @@ def main():
         "--topog-filename",
         type=str,
         default=None,
-        help="The path to a netcdf file specifying the topog. If not passed, no mask is set in the mesh.",
+        help=(
+            "Path to a netcdf file containing the topography ('topog.nc'), "
+            "used to generate the ocean land-sea mask. "
+            "If not provided, no mask will be included in the mesh.",
     )
     parser.add_argument(
         "--minimum-depth",
--- a/mesh_generation/generate_mesh.py
+++ b/mesh_generation/generate_mesh.py
@@ -48,7 +48,7 @@ from scripts_common import get_provenance_metadata, md5sum
 EARTH_R = 6.37122e6
 
 
-def mom6_mask_detection(ds, minimum_depth=0.0, masking_depth=None):
+def mom6_mask_detection(ds, minimum_depth=None, masking_depth=None):
     """
     Detect and generate an ocean mask (1 = wet, 0 = land) from a topog.nc.
     https://github.com/ACCESS-NRI/MOM6/blob/569ba3126835bfcdea5e39c46eeae01938f5413c/src/initialization/MOM_grid_initialize.F90#L1180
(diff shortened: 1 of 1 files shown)


mesh_generation/generate_mesh.py
```diff
@@ -48,6 +48,40 @@
...
+        mask = (depth > minimum_depth) & is_wet
+    else:
+        if masking_depth > minimum_depth:
+            raise ValueError(
+                "MASKING_DEPTH must be less than or equal to MINIMUM_DEPTH!"
+            )
+
+        mask = (depth > masking_depth) & is_wet
+
+    if cice_mask_filename:
```
anton-seaice, line 77: I wouldn't add the generating kmt, i'm planning to remove the need for a kmt in the future, and we already do it ([here](https://github.com/ACCESS-NRI/make_OM3_025deg_topo/blob/cadf591470293c7c379a5889380f6be4f9acd45e/gen_topo.sh#L81))

mesh_generation/generate_mesh.py
```diff
@@ -48,6 +48,40 @@
...
+
+    depth = ds["depth"]
+    print("a")
+    # topog contains nans
+    is_wet = ~np.isnan(depth)
+
+    if masking_depth is None:
+        mask = (depth > minimum_depth) & is_wet
+    else:
+        if masking_depth > minimum_depth:
```
anton-seaice, line 70: The need to use minimum_depth here is probably correct, but also messy. In other places we just assume the minimum depth is 0

e.g.
https://github.com/COSIMA/bathymetry-tools/blob/a29d07a28ce1413bcdfcdbb96356fd7a04506995/src/topography.f90#L726

Not quite sure the neatest way of handling that!

mesh_generation/generate_mesh.py
minghangli-uni: Thanks @anton-seaice , 
> i'm planning to remove the need for a kmt in the future,

When will this happen?

> and we already do it ([here](https://github.com/ACCESS-NRI/make_OM3_025deg_topo/blob/cadf591470293c7c379a5889380f6be4f9acd45e/gen_topo.sh#L81))

but this still needs the ocean mask file as an input?

anton-seaice: It does use the ocean_mask file, but it's created from the bathymetry

minghangli-uni: Ah yeah, I'll remove it

mesh_generation/generate_mesh.py
minghangli-uni: > In other places we just assume the minimum depth is 0

Then we should make a patch to `topography.f90`! 😄

minghangli-uni:
It would be good if you could have a look? @dougiesquire

anton-seaice:
This looks good @minghangli-uni - do you have example output ?

We'll need to note this weird inter-dependency on setting minimum depth consistently with MOM_input somewhere (the config docs I think ?) - can you make an issue / PR for that

minghangli-uni:
> This looks good @minghangli-uni - do you have example output ?

Ah yeah, i just regenerated those two and did a comparison. They are identical for both w/ mask and w/o mask.

`/g/data/tm70/ml0072/COMMON/git_repos/workflow_grid_mesh_generation/2025.08.08.3.25km`


> We'll need to note this weird inter-dependency on setting minimum depth consistently with MOM_input somewhere (the config docs I think ?) - can you make an issue / PR for that

I'll do that .

anton-seaice (approved):
Thanks @minghangli-uni 

Please squash merge!

Issue #80, opened 2025-08-15
Switch to explicit required flags for input/output in make_salt_sfc_restore_from_regridded_woa.py


minghangli-uni:
The current `make_salt_sfc_restore_from_regridded_woa.py` expects positional arguments, but its provenance string suggests flag-style arguments (eg --input_path, --output_path). This mismatch is confusing.

I'll open a PR for a quick fix.

PR #81, opened 2025-08-15
switch positional args to flag args in make_salt_sfc_restore_from_regridded_woa.py

minghangli-uni:
closes https://github.com/ACCESS-NRI/om3-scripts/issues/80

### DIFF
--- a/salinity_restoring_generation/make_salt_sfc_restore_from_regridded_woa.py
+++ b/salinity_restoring_generation/make_salt_sfc_restore_from_regridded_woa.py
@@ -126,13 +126,15 @@ if __name__ == "__main__":
         description="Process and concatenate NetCDF files with smoothing."
     )
     parser.add_argument(
-        "input_path",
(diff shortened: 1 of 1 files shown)


salinity_restoring_generation/make_salt_sfc_restore_from_regridded_woa.py
```diff
@@ -126,13 +126,15 @@ def main(input_path, output_path):
         description="Process and concatenate NetCDF files with smoothing."
     )
     parser.add_argument(
-        "input_path",
+        "--input_path",
```
chrisb13, line 129: So it just doesn't work without the `--`?

minghangli-uni: yes, -- is a flag argument.

anton-seaice (approved):
Looks good :)

Issue #85, opened 2025-09-19
Add Payu init script for `diag_table` generation


minghangli-uni:
When using [access-experiment-generator](https://github.com/ACCESS-NRI/access-experiment-generator), updating `diag_table` after creating separate branches with [access-experiment-runner](https://github.com/ACCESS-NRI/access-experiment-runner) is a bit cumbersome.

Currently, users need to manually run:
```
python3 make_diag_table.py
```
each time new branches are created.

Add a Payu init script that automatically runs `make_diag_table.py` after branches are created. This will streamline the workflow and remove the need for manual execution.

This feature is not intended for production use but is a convenient practice for testing and development.

PR #86, opened 2025-09-21
Add a payu init script to auto-generate updated diag_table after payu…

minghangli-uni:
closes https://github.com/ACCESS-NRI/om3-scripts/issues/85

This PR introduces a Payu init script that automates running `make_diag_table.py` when using the [experiment generator](https://github.com/ACCESS-NRI/access-experiment-generator) and [experiment runner](https://github.com/ACCESS-NRI/access-experiment-runner) workflow.

With this change, users no longer need to manually execute:
```python
python3 make_diag_table.py
```

after creating branches.

This is Intended primarily for testing and development, not for production runs.

### DIFF
--- a/payu_config/init_scripts/run_make_diag_table.sh
+++ b/payu_config/init_scripts/run_make_diag_table.sh
@@ -42,7 +42,7 @@ else
     cd $BASE
     echo "-- Cloned make_diag_table repo to $DEST"
 
-    if [ $MODEL = "access-om2" ]; then
+    if [[ $MODEL == "access-om2" || $MODEL == "access-om3" ]]; then
     python3 make_diag_table.py
     else
     cp -f diag_table_standard_source.yaml diag_table_source.yaml


payu_config/init_scripts/run_make_diag_table.sh
```diff
@@ -0,0 +1,53 @@
...
+if [ ! -f "config.yaml" ]; then
+    echo "Error: config.yaml not found!"
+    exit 1
+fi
+MODEL=$(grep '^model:' config.yaml | awk '{print $2}')
+echo "-- Detected model: $MODEL"
+
+case $MODEL in
+    access-om3)
+        DEST="diagnostic_profiles/source_yaml_files/make_diag_table"
```
aekiss: wrong path - see https://github.com/ACCESS-NRI/access-om3-configs/issues/959

payu_config/init_scripts/run_make_diag_table.sh
```diff
@@ -0,0 +1,53 @@
...
+mkdir -p $BASE
+
+rm -rf $DEST
+git clone https://github.com/COSIMA/make_diag_table $DEST
+mv $DEST/make_diag_table.py $BASE/
+
+cd $BASE
+echo "-- Cloned make_diag_table repo to $DEST"
+
+if [ $MODEL = "access-om2" ]; then
```
aekiss: or access-om3 - see https://github.com/ACCESS-NRI/access-om3-configs/issues/959

payu_config/init_scripts/run_make_diag_table.sh
```diff
@@ -0,0 +1,53 @@
...
+        exit 1
+        ;;
+esac
+
+BASE=$(dirname $DEST)
+
+mkdir -p $BASE
+
+rm -rf $DEST
+git clone https://github.com/COSIMA/make_diag_table $DEST
```
aekiss: this will update to whatever is the latest, which breaks provenance - would be better to have make_diag_table as a submodule

aekiss (commented):
I'm not sure why you'd want to do it this way. It would be neater and have proper provenance to include `make_diag_table` as a git submodule

aekiss:
or use `/g/data/vk83/apps/make_diag_table/make_diag_table.py`, though that won't retain provenance of what version of make_diag_table was used

dougiesquire:
I think we want to add `make_diag_table` as a dependency of Payu and do this from the Payu driver

minghangli-uni:
> I think we want to add make_diag_table as a dependency of Payu and do this from the Payu driver

I remember this being discussed a while ago and we were seeking approval from @aekiss on making `make_diag_table` a Payu dependency. It's good that this PR has brought the topic back up.

payu_config/init_scripts/run_make_diag_table.sh
minghangli-uni: This is done in [1c96cd1](https://github.com/ACCESS-NRI/om3-scripts/pull/86/commits/1c96cd19b82d60cd514bedaae0eef17a3e839fc9)

payu_config/init_scripts/run_make_diag_table.sh
minghangli-uni: Fixed in [b2c9bb1](https://github.com/ACCESS-NRI/om3-scripts/pull/86/commits/b2c9bb10a3f67cf19a9df70e1df79e8f9842947e)

aekiss:
I'm happy for payu to depend on make_diag_table, but it sounds more like something @aidanheerdegen or team would be better positioned to approve

aidanheerdegen:
I APPROVE!

dougiesquire:
@aekiss are you okay with me moving `make_diag_table.py` into the ACCESS-NRI Github org?

aekiss:
I suppose, if it's actually necessary

dougiesquire:
Necessary, no. Though if we are going to deploy to our conda channel it does make things a little cleaner.

We'll likely move to yaml file support in FMS with the next ACCESS-OM3 release and this may change how we want to handle these files anyway. So let's just leave in the COSIMA org for now and I'll only just the ability to install with pip.

minghangli-uni:
Thanks everyone for the discussion. Closing this PR for now. We can revisit this through a Payu-level change.

PR #100, opened 2025-12-10
use wrap-lons by default

### LINKED ISSUE #99: should --wrap-lons be the default for mesh generation
In [generate_mesh.py](https://github.com/ACCESS-NRI/om3-scripts/blob/main/mesh_generation/generate_mesh.py) it is possible to have the same points on a mesh not get connected because they use different longitude ranges.

For example, our "current" datm mesh (e.g. '/g/data/vk83/configurations/inputs/access-om3/share/meshes/share/2024.09.16/JRA55do-datm-ESMFmesh.nc') use both -0.28125 (°E) and 359.71875 for coordinates and then do not get connected.

The easy fix for this is to always use the `--wrap-lons` option when generating meshes. This means that 359.71875°E is used consistently and the mesh closes correctly east to west (see https://github.com/ACCESS-NRI/access-om3-configs/issues/968#issuecomment-3634916549)

Our other meshes all appear to use the `--wrap-lons` option already.

I don't normally favour changing defaults, but in this case, it gets rid of a gotcha we don't need!

anton-seaice:
Closes #99

### DIFF
--- a/mesh_generation/generate_mesh.py
+++ b/mesh_generation/generate_mesh.py
@@ -520,7 +520,7 @@ def main():
         help=(
             "Path to a netcdf file containing the topography ('topog.nc'), "
             "used to generate the ocean land-sea mask. "
-            "If not provided, no mask will be included in the mesh.",
+            "If not provided, no mask will be included in the mesh."
         ),
     )
     parser.add_argument(


mesh_generation/generate_mesh.py
```diff
@@ -502,7 +502,7 @@ def main():
     )
     parser.add_argument(
         "--wrap-lons",
-        default=False,
+        default=True,
         action="store_true",
```
dougiesquire: I don't think this is correct.

`action="store_true"` means that
- If the user does not provide `--wrap-lons`, the value is False.
- If the user does provide `--wrap-lons`, the value becomes True.

I think you'll actually need to change the flag, e.g. `--dont-wrap-lons`

anton-seaice:
Thanks @dougiesquire - good pickup. I changed wrap-lons to be a string which acts like a boolean

dougiesquire:
Hmmm, that won't work if people try to copy a command they've previously run that includes `--wrap-lons`, but I guess that's okay

anton-seaice:
Is this change ok? `--wrap-lons` was going to fail anyway if a `--no-wrap-lons` was added

dougiesquire (approved):
Sorry, dropped off my radar. Looks good. Just pushed a tiny change to allow `python generate_mesh.py -h` to work

Thanks @anton-seaice

PR #105, opened 2026-01-13
Revisit bottom roughness generation

minghangli-uni:
The existing bottom roughness calculation follows [Jayne & St Laurent 2001](http://dx.doi.org/10.1029/2000GL012044),

> The bottom roughness, h^2, is computed from the Smith and Sandwell [1997] 1/30° ocean topography database. Over each grid cell, a polynomial sloping surface is fit to the bottom topography (given by H = a + bx + cy + dxy), and the residual heights are used to compute h^2, the mean-square bottom roughness averaged over the grid cell", where the grid resolution is 0.5°.

but applying this approach directly on high-res ocean models is problematic. When roughness is computed at the model grid scale, the resulting h^2 becomes explicitly dependent on model resolution, which is unphysical. In the absence of explicit tidal forcing, resolved scale topography will not contribute to internal tide generation, so generation by resolved scales also needs to be parameterised. A clear summary of this issue is given by @aekiss in https://github.com/ACCESS-NRI/access-om3-configs/issues/457

Below is the key part of that summary. 

The key physical insight is that bottom roughness should instead be defined relative to the internal tide length scale, not the model grid spacing,

> roughness should instead be calculated relative to the topography smoothed over a length scale of the lowest-mode internal tide wavelength, because roughness on finer scales than this is effective in generating internal tides. This is around 180km in the deep ocean, but depends on stratification and water depth, becoming shorter in shallow water. Importantly, this smoothing scale is independent of model grid resolution.

@CallumJShakespeare has provided a matlba implementation of this approach https://github.com/CallumJShakespeare/global_tide_inputs, that workflow computes roughness using:
- internal tide length scales derived from WOA18,
- MATLAB-specific routines (e.g. `interp2`).

To integrate this method into the ACCESS-OM3 workflow and update the stratification input to WOA23, this PR re-implements the method in Python with MPI parallelism.

Hence this script does following things,

- [x] Compute the buoyancy frequency squared (N^2) from WOA23 T/S data
- [x] From N2, computes the mode-1 internal tide wavelength (lambda1) at a given tidal frequency (e.g. M2)
- [x] Smooth high-resolution bathymetry (SYNBATH) using a polar, Gaussian-weighted average with radius $\lambda1$, producing a physically meaningful mean depth.
- [x] Compute depth variance relative to this smoothed depth, getting bottom roughness independent of model grid resolution.
- [x] Interpolate the resulting roughness field onto model grids using esmf regridding.
- [x] Complete metadata and provenance.

Notes:
 - The computation involves interpolation over 670k ocean points, so a task-based MPI distribution is used. Work is distributed over valid ocean cells rather than latitude stripes. This avoids severe load imbalance due to land masks.
 - The original matlab code uses `interp2` for bathymetric sampling. But in this implementation bilinear interpolation is performed manually:
     - to minimise overhead in tight loops
     - to enforce strict NaN propagation (if any corner of the interpolation cell is nan, the result is nan), preventing land contamination of ocean roughness.

EDIT: 15 Jan 2026
 - Regridding is performed separately by `generate_bottom_roughness_regrid.py`. This is intentionally split out due to a known `xesmf + mpi4py` issue in the analysis environment: https://github.com/ACCESS-NRI/ACCESS-Analysis-Conda/issues/207
 - With 72 cpus, the full workflow completes in about 57 minutes of wall time. If the intermediate file already exists, the regridding step alone takes only about 1 minute.
 - Once the intermediate bottom-roughness file is available, regridding to different model resolutions is very fast.

### DIFF
--- a/external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
+++ b/external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
@@ -34,24 +34,23 @@
 #    - Computes the Gaussian-weighted variance of depth residuals:
 #         h^2 = <(depth - mean_depth)^2>
 #
-# 5) Regrid to the MOM6 grid
-#    - Conservatively regrids the resulting h^2 field from the regular WOA grid onto a MOM6
-#      mesh using xesmf.
+# 5) Regrid to the MOM6 grid (done separately)
+#    - Regridding is performed by generate_bottom_roughness_regrid.py.
+#    - This is intentionally split out due to a known xesmf + mpi4py issue in the analysis
+#      environment: https://github.com/ACCESS-NRI/ACCESS-Analysis-Conda/issues/207
+#    - The regrid step conservatively maps h^2 from the regular WOA grid onto a MOM6 mesh.
 #
 # Usage:
-#   mpirun -n <ranks> python3 generate_bottom_roughness.py \
+#   mpirun -n <ranks> python3 generate_bottom_roughness_intermediate_woa.py \
 #       --woa23_temp_file /path/to/woa23_temp.nc \
 #       --woa23_salt_file /path/to/woa23_salt.nc \
 #       --synbath_file    /path/to/SYNBATH.nc \
-#       --topog_file      /path/to/model_topog.nc \
-#       --hgrid_file      /path/to/hgrid.nc \
 #       --chunk_lat 800 \
 #       --chunk_lon 1600 \
 #       --nmodes 100 \
 #       --ntheta 180 \
 #       --omega 1.405189e-4 # M2 \
-#       --output_file bottom_roughness.nc \
-#       [--woa_output_file woa_intermediates.nc]
+#       --woa_intermediate_file woa_intermediates.nc
 #
 # Notes:
 # - The implementation follows the matlab reference workflow provided by Callum Shakespeare
@@ -67,7 +66,6 @@
 #   - xarray
 #   - numpy
 #   - mpi4py
-#   - xesmf
 #
 # Modules:
 #   module use /g/data/xp65/public/modules
@@ -84,12 +82,10 @@ from mpi4py import MPI
 import xarray as xr
 import gsw
 from dataclasses import dataclass
-import xesmf as xe
 
 path_root = Path(__file__).parents[1]
 sys.path.append(str(path_root))
 from scripts_common import get_provenance_metadata, md5sum
-from mesh_generation.generate_mesh import mom6_mask_detection
 
 
 def coriolis_f(lat: xr.DataArray) -> xr.DataArray:
@@ -533,129 +529,9 @@ def compute_mean_depth_and_var_points(
     return None, None
 
 
-def src_1d_corners(coord: xr.DataArray, name: str) -> xr.DataArray:
-    c = coord.values
-    mid = 0.5 * (c[1:] + c[:-1])
-    b = np.empty(c.size + 1)
-    b[1:-1] = mid
-    b[0] = c[0] - (mid[0] - c[0])
-    b[-1] = c[-1] + (c[-1] - mid[-1])
-
-    return xr.DataArray(b, dims=(f"{name}_b",), name=f"{name}_b")
-
-
-def build_source_ds(
-    depth_var: xr.DataArray, lon: xr.DataArray, lat: xr.DataArray
-) -> xr.Dataset:
-    """
-    Build xesmf source dataset from depth_var with lon/lat coords.
-    """
-    lon_b_1d = src_1d_corners(lon, "lon")
-    lat_b_1d = src_1d_corners(lat, "lat")
-    lat_b2d, lon_b2d = xr.broadcast(lat_b_1d, lon_b_1d)
-
-    source_ds = xr.Dataset(
-        data_vars={"mask": (("lat", "lon"), depth_var.values)},
-        coords={
-            "lon": lon,
-            "lat": lat,
-            "lon_b": (("lat_b", "lon_b"), lon_b2d.values),
-            "lat_b": (("lat_b", "lon_b"), lat_b2d.values),
-        },
-    )
-    return source_ds
-
-
-def build_target_ds(
-    mask: xr.DataArray,
-    hgrid_x: xr.DataArray,
-    hgrid_y: xr.DataArray,
-    hgrid_xc: xr.DataArray,
-    hgrid_yc: xr.DataArray,
-) -> xr.Dataset:
-    """
-    Build xesmf target dataset from MOM6 hgrid coords and mask.
-    """
-    target_ds = xr.Dataset(
-        data_vars={"mask": (("y", "x"), mask)},
-        coords={
-            "lon": (("y", "x"), hgrid_x.values),
-            "lat": (("y", "x"), hgrid_y.values),
-            "lon_b": (("y_b", "x_b"), hgrid_xc.values),
-            "lat_b": (("y_b", "x_b"), hgrid_yc.values),
-        },
-    )
-    return target_ds
-
-
-def regrid_depth_var_to_mom6(
-    depth_var: xr.Dataset,
-    topog_file: str,
-    hgrid_file: str,
-    method: str = "conservative_normed",
-    periodic: bool = True,
-) -> xr.Dataset:
-    """
-    Regrid depth_var (on regular WOA lon/lat grid) onto MOM6 grid using xESMF.
-    """
-
-    source_lon = depth_var["lon"]
-    source_lat = depth_var["lat"]
-
-    # Load model topog file then generate ocean mask ranging from [-280, 80]
-    topog = xr.open_dataset(topog_file)
-    mask = mom6_mask_detection(topog)
-
-    # model hgrid
-    hgrid = xr.open_dataset(hgrid_file)
-
-    # match your slicing exactly
-    hgrid_x = hgrid.x[1::2, 1::2]
-    hgrid_y = hgrid.y[1::2, 1::2]
-    hgrid_xc = hgrid.x[::2, ::2]
-    hgrid_yc = hgrid.y[::2, ::2]
-
-    source_ds = build_source_ds(depth_var, lon=source_lon, lat=source_lat)
-    target_ds = build_target_ds(mask, hgrid_x, hgrid_y, hgrid_xc, hgrid_yc)
-
-    regridder_kwargs = dict(
-        method=method,
-        periodic=periodic,
-    )
-
-    regridder = xe.Regridder(source_ds, target_ds, **regridder_kwargs)
-
-    target_ds["h2"] = regridder(depth_var)
-    target_ds["h2"] = target_ds["h2"].fillna(0.0)
-
-    # tidy up attrs
-    target_ds = target_ds.drop_vars(["lon_b", "lat_b", "mask"])
-
-    tmp_attrs = {
-        "lon": {
-            "long_name": "Longitude",
-            "units": "degrees_east",
-        },
-        "lat": {
-            "long_name": "Latitude",
-            "units": "degrees_north",
-        },
-        "h2": {
-            "long_name": "Bottom roughness squared (h^2) for internal tide generation",
-            "units": "m^2",
-            "regrid_method": "conservative_normed",
-        },
-    }
-
-    for var, attrs in tmp_attrs.items():
-        target_ds[var].attrs.update(attrs)
-
-    return target_ds
-
-
 def main():
     parser = argparse.ArgumentParser(
-        description="Compute mean depth based on lambda1 computed from WOA23."
+        description="Compute depth variance based on lambda1 computed from WOA23."
     )
     parser.add_argument(
         "--woa23_temp_file",
@@ -706,28 +582,10 @@ def main():
         help="Tidal frequency in rad/s (default M2)",
     )
     parser.add_argument(
-        "--topog_file",
-        type=str,
-        required=True,
-        help="Path to the model bathymetry file.",
-    )
-    parser.add_argument(
-        "--hgrid_file",
-        type=str,
-        required=True,
-        help="Path to the model hgrid file.",
-    )
-    parser.add_argument(
-        "--woa_output_file",
+        "--woa_intermediate_file",
         type=str,
         default=None,
-        help="Optional path to the raw output file including lambda1, mean_depth, depth_var on WOA grid.",
-    )
-    parser.add_argument(
-        "--output_file",
-        type=str,
-        default="bottom_roughness.nc",
-        help="Path to the bottom roughness output file.",
+        help="Intermediate output file including lambda1, mean_depth, depth_var on WOA grid.",
     )
     args = parser.parse_args()
 
@@ -867,18 +725,8 @@ def main():
                 "omega_rad_s": args.omega,
             },
         )
-        if args.woa_output_file is not None:
-            ds_woa_output.to_netcdf(args.woa_output_file)
-            print(f"Output written to {args.woa_output_file}")
-
-        # Regridding to model grid
-        regrid_depth_var = regrid_depth_var_to_mom6(
-            depth_var=ds_woa_output["depth_var"],
-            topog_file=args.topog_file,
-            hgrid_file=args.hgrid_file,
-            method="conservative_normed",
-            periodic=True,
-        )
+        ds_woa_output.to_netcdf(args.woa_intermediate_file)
+        print(f"Output written to {args.woa_intermediate_file}")
 
         # Add provenance metadata and MD5 hashes for input files.
         this_file = os.path.normpath(__file__)
@@ -887,8 +735,6 @@ def main():
             f"--woa23_temp_file={args.woa23_temp_file} "
             f"--woa23_salt_file={args.woa23_salt_file} "
             f"--synbath_file={args.synbath_file} "
-            f"--topog-file={args.topog_file} "
-            f"--hgrid-file={args.hgrid_file} "
             f"--chunk-lat={args.chunk_lat} "
             f"--chunk-lon={args.chunk_lon} "
             f"--nmodes={args.nmodes} "
@@ -896,8 +742,7 @@ def main():
             f"--earth-radius={args.earth_radius} "
             f"--omega={args.omega} "
             f"--print-every={args.print_every} "
-            f"--woa-output-file={args.woa_output_file} "
-            f"--output_file={args.output_file}"
+            f"--woa-intermediate-file={args.woa_intermediate_file} "
         )
 
         history = get_provenance_metadata(this_file, runcmd)
@@ -906,14 +751,12 @@ def main():
             f"{args.woa23_temp_file} (md5 hash: {md5sum(args.woa23_temp_file)})",
             f"{args.woa23_salt_file} (md5 hash: {md5sum(args.woa23_salt_file)})",
             f"{args.synbath_file} (md5 hash: {md5sum(args.synbath_file)})",
-            f"{args.hgrid_file} (md5 hash: {md5sum(args.hgrid_file)})",
-            f"{args.topog_file} (md5 hash: {md5sum(args.topog_file)})",
         ]
         global_attrs["inputFile"] = ", ".join(file_hashes)
-        regrid_depth_var.attrs.update(global_attrs)
+        ds_woa_output.attrs.update(global_attrs)
 
-        regrid_depth_var.to_netcdf(args.output_file)
-        print(f"Output written to {args.output_file}")
+        ds_woa_output.to_netcdf(args.woa_intermediate_file)
+        print(f"Output written to {args.woa_intermediate_file}")
 
 
 if __name__ == "__main__":
--- a/external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
+++ b/external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
@@ -531,7 +531,7 @@ def compute_mean_depth_and_var_points(
 
 def main():
     parser = argparse.ArgumentParser(
-        description="Compute depth variance based on lambda1 computed from WOA23."
+        description="Compute a grid-independent bottom roughness by computing N^2 from WOA23 T/S, estimating a mode-1 internal-tide wavelength, and using it to smooth high-resolution bathymetry before evaluating depth variance."
     )
     parser.add_argument(
         "--woa23_temp_file",
--- a/external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
+++ b/external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
@@ -381,6 +381,112 @@ def gatherv_indexed(
     return out1d
 
 
+def fill_missing_data_laplace(
+    field: np.ndarray, mask: np.ndarray, periodic_lon: bool = True
+) -> np.ndarray:
+    """
+    Fill nans smoothly by solving a discrete Laplace problem over the wet domain.
+
+    This is adapted from https://github.com/ACCESS-NRI/om3-scripts/blob/main/chlorophyll/chl_climatology_and_fill.py,
+    which originally adapted from https://github.com/adcroft/interp_and_fill/blob/main/Interpolate%20and%20fill%20SeaWIFS.ipynb
+
+    This implementation otherwise assumes a regular lat/lon grid tri-polar (WOA),
+    hence tripolar topology is intentionally not handled here.
+
+    Periodic boundary conditions are supported in longitude only (global configuration).
+    For regional configurations, set periodic_lon=False is not implemented yet.
+    """
+    nj, ni = field.shape
+    # Find the missing points to fill (nan in field but mask > 0)
+    missing_mask = np.isnan(field) & (mask > 0)
+    if not np.any(missing_mask):
+        # no missing data to fill but also guarantee nans on dry cells
+        return np.where(mask > 0, field, np.nan)
+
+    # change nan to 0 for the sparse matrix construction
+    work = np.where(np.isnan(field), 0.0, field)
+    missing_j, missing_i = np.where(missing_mask)
+    n_missing = missing_j.size
+    ind = np.full((nj, ni), -1, dtype=np.int64)
+    ind[missing_j, missing_i] = np.arange(n_missing)
+
+    # Sparse matrix
+    A = sp.lil_matrix((n_missing, n_missing))
+    b = np.zeros(n_missing)
+    ld = np.zeros(n_missing)
+
+    def _process_neighbour(n: int, jn: int, in_: int) -> None:
+        """Process neighbour at (jn, in_) for row n."""
+        if mask[jn, in_] <= 0:
+            return
+
+        ld[n] -= 1
+        idx = ind[jn, in_]
+
+        if idx >= 0:
+            A[n, idx] = 1
+        else:
+            b[n] -= work[jn, in_]
+
+    for n in range(n_missing):
+        j = missing_j[n]
+        i = missing_i[n]
+
+        if periodic_lon:
+            im1 = (i - 1) % ni  # west
+            ip1 = (i + 1) % ni  # east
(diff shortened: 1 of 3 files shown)


minghangli-uni:
image

Overall the new \sqrt(h^2) (bottom left) show strong qualitative agreement in the large-scale features seen in both @CallumJShakespeare’s 100 km config (top left) and the OM2 configuration (top right). The peak magnitudes in the new field are comparable to those in Callum’s matlab result, but are roughly twice those in OM2, which is [capped at 1000m ](https://github.com/ACCESS-NRI/om3-scripts/pull/53#:~:text=The%20OM2%20roughness%20field%20is%20capped%20at%20exactly%201000%20m%20(this%20seems%20a%20bit%20suspicious%20to%20me%20and%20suggests%20that%20some%20postprocessing%20might%20have%20been%20applied)%2C). The bottom-right figure shows the roughness field currently used in OM3.

@aekiss @CallumJShakespeare would you mind taking a look and letting me know if you have any suggestions?

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,763 @@
...
+    if rank == 0:
+        depth_var = depth_var_1d.reshape(nj, ni)
+        return mean_depth, depth_var
+
+    return None, None
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Compute depth variance based on lambda1 computed from WOA23."
```
minghangli-uni:
```suggestion
        description="Compute a grid-independent bottom roughness by computing N^2 from WOA23 T/S, estimating a mode-1 internal-tide wavelength, and using it to smooth high-resolution bathymetry before evaluating depth variance."
```

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,763 @@
...
+#    - Re-samples synbath on the same polar stencil.
+#    - Interpolates the smoothed mean depth to the stencil points.
+#    - Computes the Gaussian-weighted variance of depth residuals:
+#         h^2 = <(depth - mean_depth)^2>
+#
+# 5) Regrid to the MOM6 grid (done separately)
+#    - Regridding is performed by generate_bottom_roughness_regrid.py.
+#    - This is intentionally split out due to a known xesmf + mpi4py issue in the analysis
+#      environment: https://github.com/ACCESS-NRI/ACCESS-Analysis-Conda/issues/207
+#    - The regrid step conservatively maps h^2 from the regular WOA grid onto a MOM6 mesh.
```
aekiss: I don't see a good reason to map conservatively. Something smooth would be better, e.g. bilinear or patch.

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,763 @@
...
+#
+# Usage:
+#   mpirun -n <ranks> python3 generate_bottom_roughness_intermediate_woa.py \
+#       --woa23_temp_file /path/to/woa23_temp.nc \
+#       --woa23_salt_file /path/to/woa23_salt.nc \
```
aekiss: It might not make much difference, but would be better to use the annual-mean WOA "decav" files, rather than the January WOA "decav" files used for the initial condition.

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,763 @@
...
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Compute a grid-independent bottom roughness by computing N^2 from WOA23 T/S, estimating a mode-1 internal-tide wavelength, and using it to smooth high-resolution bathymetry before evaluating depth variance."
+    )
+    parser.add_argument(
+        "--woa23_temp_file",
+        type=str,
+        required=True,
+        help="Path to WOA23 temperature file.",
```
aekiss:
```suggestion
        help="Path to WOA temperature file (annual mean recommended).",
```

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,763 @@
...
+    )
+    parser.add_argument(
+        "--woa23_salt_file",
+        type=str,
+        required=True,
+        help="Path to WOA23 salinity file.",
```
aekiss:
```suggestion
        help="Path to WOA salinity file (annual mean recommended).",
```

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,763 @@
...
+        description="Compute a grid-independent bottom roughness by computing N^2 from WOA23 T/S, estimating a mode-1 internal-tide wavelength, and using it to smooth high-resolution bathymetry before evaluating depth variance."
+    )
+    parser.add_argument(
+        "--woa23_temp_file",
+        type=str,
+        required=True,
+        help="Path to WOA23 temperature file.",
+    )
+    parser.add_argument(
+        "--woa23_salt_file",
```
aekiss:
```suggestion
        "--woa_salt_file",
```

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,763 @@
...
+        )
+        ds_woa_output.to_netcdf(args.woa_intermediate_file)
+        print(f"Output written to {args.woa_intermediate_file}")
+
+        # Add provenance metadata and MD5 hashes for input files.
+        this_file = os.path.normpath(__file__)
+        runcmd = (
+            f"mpirun -n $PBS_NCPUS python3 {os.path.basename(this_file)} "
+            f"--woa23_temp_file={args.woa23_temp_file} "
+            f"--woa23_salt_file={args.woa23_salt_file} "
```
aekiss:
```suggestion
            f"--woa_temp_file={args.woa_temp_file} "
            f"--woa_salt_file={args.woa_salt_file} "
```

external_tidal_generation/generate_bottom_roughness_regrid.py
```diff
@@ -0,0 +1,255 @@
+#!/usr/bin/env python3
+# Copyright 2026 ACCESS-NRI and contributors.
+# SPDX-License-Identifier: Apache-2.0
+
+# =========================================================================================
+# Bottom roughness regridding for internal-tide generation (h^2)
+#
+# This script regrids a precomputed bottom-roughness field (h^2), generated on the
+# regular WOA23 lat-lon grid onto a MOM6 model grid using xesmf.
```
aekiss:
```suggestion
# regular WOA lat-lon grid onto a MOM6 model grid using xesmf.
```

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,763 @@
...
+#    - The regrid step conservatively maps h^2 from the regular WOA grid onto a MOM6 mesh.
+#
+# Usage:
+#   mpirun -n <ranks> python3 generate_bottom_roughness_intermediate_woa.py \
+#       --woa23_temp_file /path/to/woa23_temp.nc \
+#       --woa23_salt_file /path/to/woa23_salt.nc \
+#       --synbath_file    /path/to/SYNBATH.nc \
+#       --chunk_lat 800 \
+#       --chunk_lon 1600 \
+#       --nmodes 100 \
```
aekiss:
```suggestion
#       --nradial 100 \
```

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,763 @@
...
+
+    r: np.ndarray
+    cos_t: np.ndarray
+    sin_t: np.ndarray
+    weight: np.ndarray
+    weight_sum: float
+
+    @classmethod
+    def build(cls, nmodes: int, ntheta: int):
+        nr = 2 * nmodes + 1
```
aekiss:
```suggestion
    def build(cls, nradial: int, ntheta: int):
        nr = 2 * nradial + 1
```

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,763 @@
...
+    tasks = comm.bcast(tasks, root=0)
+
+    local_tasks = tasks[rank::size]
+    n_local = local_tasks.size
+    if rank == 0:
+        print(f"avg tasks per rank ~ {tasks.size/size:.1f}")
+    print(f"[Rank {rank}] local tasks = {n_local}")
+
+    # Precompute polar weights
+    polar = PolarWeights.build(nmodes, ntheta)
```
aekiss:
```suggestion
    polar = PolarWeights.build(nradial, ntheta)
```

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,763 @@
...
+    r: np.ndarray
+    cos_t: np.ndarray
+    sin_t: np.ndarray
+    weight: np.ndarray
+    weight_sum: float
+
+    @classmethod
+    def build(cls, nmodes: int, ntheta: int):
+        nr = 2 * nmodes + 1
+        r = np.linspace(-1, 1, nr)
```
aekiss: I'm confused by this line. theta runs from 0 to 2pi, so if r runs from -1 to 1 it covers the disc twice.

I think we can do this instead, right? If so we can scrap the line above that defines nr, and replace nr with nradial+1 in line 148
```suggestion
        r = np.linspace(0, 1, nradial+1)
```

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,763 @@
...
+        theta = np.linspace(0, 2 * np.pi, ntheta, endpoint=False)
+
+        cos_t = np.cos(theta)[None, :]
+        sin_t = np.sin(theta)[None, :]
+        r_col = r[:, None]
+
+        weight = np.exp(-((2 * r_col) ** 2)) * (2 * np.pi) * np.abs(r_col)
```
aekiss:
```suggestion
        weight = np.exp(-((2 * r_col) ** 2))
```

I'm confused by this line (which matches @callumjshakespeare's code) - why is it multiplied by abs(r_col)? That's a radially linear weight of a Gaussian, so sets the central weight to zero, which isn't what we want.

We could also drop the factor of `(2 * np.pi)`, as it's divided out when we divide by `weight_sum`.

aekiss: Note this is actually a cropped Gaussian, as r is at most 1. So `np.exp(-((2 * r_col) ** 2))` is at least `np.exp(-(2 ** 2))` = 0.018. But that's close enough to zero I think.

aekiss: Oh never mind, my mistake, the weight is including a factor of the element area in polar coords.

minghangli-uni: Yup correct. But I think it shouldn't have a meaningful impact on the results

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,764 @@
...
+            p3d,
+            lat3,
+        )
+
+        depth_mid_values = 0.5 * (depth.values[:-1] + depth.values[1:])
+        depth_mid = xr.DataArray(
+            depth_mid_values,
+            dims=("depth_mid",),
+            coords={"depth_mid": depth_mid_values},
+            name="depth_mid",
```
aekiss: add something lke this?
```
            attrs={"units": "m", "long_name": "Depth midway between WOA depths (positive down)"}
```

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,764 @@
...
+        )
+
+        print("computing lambda1")
+        da_lambda1 = compute_lambda(N2_da, lat, depth, args.omega)
+
+        lon_np = lon.values
+        lat_np = lat.values
+        lambda1_np = da_lambda1.values
+
+        print("loading high-res topo")
```
aekiss:
```suggestion
        print(f"loading high-res topo from {args.synbath_file}")
```

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
minghangli-uni: Good catch! In @CallumJShakespeare's code, they restrict theta to [0, pi) instead of 2pi. I would also prefer to go this way cuz it is capable of minimising risk of subtle differences near coasts

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,764 @@
...
+        "--nmodes", type=int, default=100, help="Number of modes for polar weights."
+    )
+    parser.add_argument(
+        "--ntheta",
+        type=int,
+        default=180,
+        help="Number of theta divisions for polar weights.",
+    )
+    parser.add_argument(
+        "--earth_radius", type=float, default=6371000.0, help="Earth radius in meters."
```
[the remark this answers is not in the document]
aekiss: (not that it matters much)

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
minghangli-uni: Just checked https://xesmf.readthedocs.io/en/latest/notebooks/Masking.html

> [results with other methods (e.g. bilinear) may not give masks consistent with the coarse grid.](https://xesmf.readthedocs.io/en/latest/notebooks/Masking.html#:~:text=results%20with%20other%20methods%20(e.g.%20bilinear)%20may%20not%20give%20masks%20consistent%20with%20the%20coarse%20grid.)

To regrid to a coarse grid, it is [suggested by the author](https://github.com/pangeo-data/xESMF/issues/256#issuecomment-1504413575) to use `conservative_normed`.

fyi our current default is already `conservative_normed`. But I'll still double check how it looks on the 100km, 25km, and8 km grids.

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
minghangli-uni: It is fixed in [2c65d61](https://github.com/ACCESS-NRI/om3-scripts/pull/105/commits/2c65d612cd3d20e9cfd0a8743034dacbbac334aa)

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
aekiss: ... but it would be good to avoid that bilinear mask issue when regridding to a coarser grid. In that case we don't need a smooth method, so I'm happy with `conservative_normed`.

aekiss: Is the mask issue impossible when regridding to a finer grid? If so, I'd advocate `bilinear` in that case.

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,770 @@
...
+    field: np.ndarray,
+    lon0: float,
+    lat0: float,
+    dres: float,
+    lon_x: np.ndarray,
+    lat_y: np.ndarray,
+) -> np.ndarray:
+    """
+    Bilinear interpolation on regular grid - if any corner is NaN -> output NaN.
+    """
```
aekiss: comments to explain the inputs and outputs would be helpful here

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,770 @@
...
+    def build(cls, nradial: int, ntheta: int):
+        nr = 2 * nradial + 1
+        r = np.linspace(-1, 1, nr)
+        theta = np.linspace(0, np.pi, ntheta, endpoint=False)
+
+        cos_t = np.cos(theta)[None, :]
+        sin_t = np.sin(theta)[None, :]
+        r_col = r[:, None]
+
+        weight = np.exp(-((2 * r_col) ** 2)) * np.pi * np.abs(r_col)
```
aekiss: Is `weight` supposed to be the Gaussian multiplied by the area of the cell surrounding each sampling point? If so, it's missing a factor of the radial increment times the theta increment. But I guess these divide out when the weighting is applied and divided by `weight_sum`?

aekiss: `weight` is always divided by `weight_sum` when it's applied, so why not divide `weight` by `weight_sum` here, to avoid having to pass `weight_sum` separately?

minghangli-uni: I recall I did this was to make the normalisation explicitly at the point of use. This makes it a bit clearer when a weighted average is required. But yes numerically they are equivalent.

external_tidal_generation/generate_bottom_roughness_regrid.py
```diff
@@ -0,0 +1,255 @@
...
+    hgrid_y = hgrid.y[1::2, 1::2]
+    hgrid_xc = hgrid.x[::2, ::2]
+    hgrid_yc = hgrid.y[::2, ::2]
+
+    source_ds = build_source_ds(depth_var, lon=source_lon, lat=source_lat)
+    target_ds = build_target_ds(mask, hgrid_x, hgrid_y, hgrid_xc, hgrid_yc)
+
+    regridder_kwargs = dict(
+        method=method,
+        periodic=periodic,
```
aekiss: Is this periodic in y as well as x?

Can the tripolar seam be handled somehow?

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,770 @@
...
+    r: np.ndarray
+    cos_t: np.ndarray
+    sin_t: np.ndarray
+    weight: np.ndarray
+    weight_sum: float
+
+    @classmethod
+    def build(cls, nradial: int, ntheta: int):
+        nr = 2 * nradial + 1
+        r = np.linspace(-1, 1, nr)
```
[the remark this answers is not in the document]
aekiss: I agree the compute saving are negligible, but the existing code weights the central point by zero area, despite it corresponding to a disc at the origin of radius dr/2 (or actually ntheta pizza slices of that disc, one for each of the ntheta duplicates). This completely neglects the data at the origin, which seems incorrect to me as this is actually the most relevant data.

My suggested change is a simple way to fix that. If we consider each sample to correspond to a patch of radial extent dr/2 either side of the central point, the innermost ring will fill in this central hole.

Otherwise the smoothed field will be like the bokeh of a catadioptric lens
https://en.wikipedia.org/wiki/Catadioptric_system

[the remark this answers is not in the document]
aekiss: https://en.wikipedia.org/wiki/Catadioptric_system#/media/File:Bahnhof_Dortmund_Brackel_Spiegeltele_DSC02359_smial_wp.jpg

external_tidal_generation/generate_bottom_roughness_regrid.py
```diff
@@ -0,0 +1,255 @@
...
+    )
+
+    regridder = xe.Regridder(source_ds, target_ds, **regridder_kwargs)
+
+    target_ds["h2"] = regridder(depth_var)
+    target_ds["h2"] = target_ds["h2"].fillna(0.0)
```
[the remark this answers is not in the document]
aekiss: Actually, it would be best to fill nans with extrapolated values in the intermediate file, so it only needs to be done once.

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
minghangli-uni: Added in [65e174d](https://github.com/ACCESS-NRI/om3-scripts/pull/105/commits/65e174d8d3000c3b51be500281a61b15675cedaf)

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,770 @@
...
+
+    # Now broadcast mean_depth to all ranks
+    mean_depth = comm.bcast(mean_depth, root=0)
+
+    # Now for depth variance
+    local_idx_var_depth = np.empty(n_local, dtype=np.int64)
+    local_val_var_depth = np.empty(n_local, dtype=np.float64)
+    local_val_var_depth.fill(np.nan)
+
+    for n, idx1d in enumerate(local_tasks):
```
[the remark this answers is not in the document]
minghangli-uni: I've done a small refactor in [2bfb5d4](https://github.com/ACCESS-NRI/om3-scripts/pull/105/commits/2bfb5d4e84e2dd72db50adfad3757680125236b8)

It doesnt merge the two loops, but factors out the shared polar sampling logic to improve maintainability.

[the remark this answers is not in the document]
aekiss: ok thanks

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
minghangli-uni: > I'm happy to switch the default but I'll do some initial tests first.

@aekiss  I'll hold off on the initial test for now and run it once all the comments are addressed.

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,805 @@
...
+    j1 = int(np.ceil((lat_max - synbath_lat0) / synbath_res))
+
+    h_patch = read_lonbuffer_patch(topog, j0, j1, i0_buf, i1_buf, synbath_nlon)
+    if h_patch is None:
+        return None
+
+    lon_patch0 = synbath_lon_buf0 + i0_buf * synbath_res
+    lat_patch0 = synbath_lat0 + j0 * synbath_res
+
+    depth = bilinear_interp(h_patch, lon_patch0, lat_patch0, synbath_res, lon_x, lat_y)
```
aekiss:
```suggestion
    # depth is synbath interpolated onto lon_x, lat_y.
    # NB: depth is nan at points on land or beyond the northern (or southern) edge of the synbath grid
    depth = bilinear_interp(h_patch, lon_patch0, lat_patch0, synbath_res, lon_x, lat_y)
```

external_tidal_generation/generate_bottom_roughness_regrid.py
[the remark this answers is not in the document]
aekiss: you could see what was done for chlorophyll or initial conditions

external_tidal_generation/generate_bottom_roughness_regrid.py
minghangli-uni: I just had a quick look. Although the doc doesnt explicitly mention tripolar grids, they still just a curvilinear lon/lat grid, which is supported by xesmf. Also using MOM6 hgrid as the target is how we represent the tripolar grid to xesmf so that gives esmf enough info to build weights on the actual grid geometry. So I think the approximation is acceptable for this use case?

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,770 @@
...
+        lon_patch0 = synbath_lon_buf0 + i0_buf * synbath_res
+        lat_patch0 = synbath_lat0 + j0 * synbath_res
+
+        depth = bilinear_interp(
+            h_patch, lon_patch0, lat_patch0, synbath_res, lon_x, lat_y
+        )
+
+        num = np.nansum(depth * polar.weight)
+        if np.isfinite(num):
+            local_val_mean_depth[n] = -num / polar.weight_sum
```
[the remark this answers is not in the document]
aekiss: I think something like this will fix the issue (not tested!). It actually doesn't require the definition of polar.weight to include a factor of dr*dtheta, but including that factor would make the code clearer. With this change we no longer need `polar.weight_sum`.
```suggestion
            masked_weight = np.where(np.isfinite(depth), polar.weight, 0.0)
            masked_weight_sum = np.sum(masked_weight)
            local_val_mean_depth[n] = -num / masked_weight_sum
```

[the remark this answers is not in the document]
aekiss: thanks Minghang

> there might be conditions when masked_weight_sum  is zero. That would happen if all stencil points are nans

can that happen even within this if-block?

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
```diff
@@ -0,0 +1,931 @@
...
+
+
+def fill_missing_data_laplace(
+    field: np.ndarray, mask: np.ndarray, periodic_lon: bool = True
+) -> np.ndarray:
+    """
+    Fill nans smoothly by solving a discrete Laplace problem over the wet domain.
+
+    This is adapted from https://github.com/ACCESS-NRI/om3-scripts/blob/main/chlorophyll/chl_climatology_and_fill.py,
+    which originally adapted from https://github.com/adcroft/interp_and_fill/blob/main/Interpolate%20and%20fill%20SeaWIFS.ipynb
```
aekiss: including commit hashes would be better here

external_tidal_generation/generate_bottom_roughness_regrid.py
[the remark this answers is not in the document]
aekiss: Great, thanks @minghangli-uni

It was unclear to me in a quick read, but will this fill every cell in the intermediate file with interpolated data, so there are no masked/nan cells anywhere in that file?

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
minghangli-uni: `weight_sum` now is removed in [8ca2ebc](https://github.com/ACCESS-NRI/om3-scripts/pull/105/commits/8ca2ebcc72e07c093c7cde6aae881134e79f3a12)

minghangli-uni:
> Revised patch seems to have problems, but we could use bilinear instead.

Correct. Here is what I see from the mask check.

```python
# da1: conservative_normed regridding method (initial try)
# da2: conservative_normed regridding method (revised)
# da3: bilinear regridding method (revised)
# da4: patch regridding method (revised)

for i, da in enumerate([da1, da2, da3, da4], start=1):
    same_mask = np.array_equal(np.isfinite(da.values), mask.values)
    print(f"da{i} uses same mask:", same_mask)
da1 uses same mask: False
da2 uses same mask: True
da3 uses same mask: True
da4 uses same mask: False
```

So for lower resolution configs e.g. 100km configs, we will use the `conservative_normed` method. For moderate to high resolution configs, `bilinear` interpolation will be used.

If you're happy with these results, would you be able to have a final look and approve this @aekiss ?

external_tidal_generation/generate_bottom_roughness_intermediate_woa.py
minghangli-uni: Addressed in [29a4fc1](https://github.com/ACCESS-NRI/om3-scripts/pull/105/commits/29a4fc156083aa91e0d9256514e6ead5bb19cd85)

aekiss:
There are 2 unresolved comments - let me know when they're done or rejected

aekiss (approved):
Great work, thanks @minghangli-uni !

external_tidal_generation/generate_bottom_roughness_regrid.py
aekiss: yep

PR #107, opened 2026-01-15
Map runoff to coastal cells only

### LINKED ISSUE #106: Put liquid runoff in coastal cells
Per 

https://github.com/ACCESS-NRI/access-om3-configs/issues/727

adjust `generatore_rof_weights.py` to only put runoff at coast

anton-seaice:
This change restricts target cells for runoff to those along the coast only discussed in https://github.com/ACCESS-NRI/access-om3-configs/issues/727. - Closes #106 

This is some output of friver with a remapping file (/g/data/tm70/as2285/making_inputs/om3-scripts/mesh_generation/access-om3-25km-rof-remap-weights.nc)  generated through this change 

image

There's is no no spreading of runoff away from coastal cells due to remapping from a lower resolution JRA grid to a higher resolution model grid.

The generation of the weights now is much slower, I think this file took ~1.5 hours to generate.

### DIFF
--- a/mesh_generation/generate_rof_weights.py
+++ b/mesh_generation/generate_rof_weights.py
@@ -51,8 +51,8 @@ COMP_ENCODING = {"complevel": 1, "compression": "zlib"}  # compression settings 
 def drof_remapping_weights(mesh_filename, weights_filename, nx, ny, global_attrs=None):
     # We need to generate remapping weights for use in the mediator, such that the overall volume of runoff is conserved
     # and no runoff is mapped onto land cells. Inside the mediator, the grid doesn't change as we run the mediator with
-    # the ocean grid (the DROF component does the remapping from JRA grid to mediator grid). There we use the
-    # same _mesh_file for the input and output mesh, however this same routine would work for differing input and
+    # the ocean grid (the DROF component does the remapping from JRA grid to mediator grid). Therefore we use the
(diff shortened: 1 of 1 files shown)


dougiesquire (commented):
I've had issues in the past with dilation removing narrow channels (so one would end up with no runoff). Did you notice this at all?

It's probably not an issue with a 1x1 structure element, but just thought I'd ask.

anton-seaice:
I poked around for some narrow channels and  it looks ok:

old on the left, new on the right

image

anton-seaice:
The BallTree uses a Haversine distance rather than Euclidean to account for the earth being a globe. 

I think EDT might be faster because it assumes a regular grid ? Which is close but not totally true for our grids

dougiesquire:
> The BallTree uses a Haversine distance rather than Euclidean to account for the earth being a globe.
> 
> I think EDT might be faster because it assumes a regular grid ? Which is close but not totally true for our grids

Oh, I'm not suggesting using edt for the BallTree part, just for finding the coastal cells as [above](https://github.com/ACCESS-NRI/om3-scripts/pull/107#issuecomment-3770474782). I used it here because it is consistent with the dilation approach, which just dilates by n cells, rather than any sort of real distance.

anton-seaice:
Oh the slow part is making the balltree, not finding the coastal cells

PR #111, opened 2026-01-28
Use _FillValue from source data, or sensible default

anton-seaice:
Related to https://github.com/ACCESS-NRI/access-om3-configs/issues/1095

This change sets _FillValue correctly in the RYF forcing data

### DIFF
--- a/make_ryf/make_ryf.py
+++ b/make_ryf/make_ryf.py
@@ -13,7 +13,7 @@
 #
 # Usage:
 # First start up an interactive job or an ARE session on Gadi to get enough memory:
-# qsub -I -q express -l mem=32GB -l storage=gdata/xp65+gdata/qv56+gdata/tm70+gdata/ua8 -l wd
+# qsub -I -q express -l mem=32GB -l storage=gdata/xp65+gdata/qv56+gdata/tm70 -l wd
 #
 # Then run the following to create the May-May repeat year forcings
 # module use /g/data/xp65/public/modules ; module load conda/analysis3
@@ -156,10 +156,10 @@ for year1 in years:
             # Have to give all variables a FillValue attribute, otherwise xarray
             # makes it NaN which causes floating point errors
             # copy FillValue if it exists, otherwise use default
+            source_ds = ds[baseyear]
             ryf[varname].encoding["_FillValue"] = (
-                ds[varname].encoding.get("_FillValue", FILLVALUE)
-                if varname in ds
-                else FILLVALUE
+                source_ds[varname].encoding.get("_FillValue")
+                or FILLVALUE
             )
 
             # Only process variables with 3 or more dimensions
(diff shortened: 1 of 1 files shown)


anton-seaice:
Example data in :

/g/data/tm70/as2285/making_inputs/om3-scripts/make_ryf/

Repro test in https://github.com/ACCESS-NRI/access-om3-configs/pull/1096

make_ryf/make_ryf.py
```diff
@@ -13,7 +13,7 @@
 #
 # Usage:
 # First start up an interactive job or an ARE session on Gadi to get enough memory:
-# qsub -I -q express -l mem=32GB -l storage=gdata/xp65+gdata/qv56+gdata/tm70+gdata/ua8
+# qsub -I -q express -l mem=32GB -l storage=gdata/xp65+gdata/qv56+gdata/tm70+gdata/ua8 -l wd
```
minghangli-uni:
```suggestion
# qsub -I -q express -l mem=32GB -l storage=gdata/xp65+gdata/qv56+gdata/tm70 -l wd
```

minghangli-uni (commented):
`ua8` is being decommissioned soon, so it probably makes sense to remove these now?

make_ryf/make_ryf.py
```diff
@@ -151,10 +153,14 @@
...
-            if "_FillValue" not in ryf[varname].encoding:
-                ryf[varname].encoding["_FillValue"] = FillValue
+            # Have to give all variables a FillValue attribute, otherwise xarray
+            # makes it NaN which causes floating point errors
+            # copy FillValue if it exists, otherwise use default
+            ryf[varname].encoding["_FillValue"] = (
+                ds[varname].encoding.get("_FillValue", FILLVALUE)
+                if varname in ds
+                else FILLVALUE
+            )
```
minghangli-uni: `ds` is a dict with keys of years, e.g., 1990 and 1991, so `ds[varname] if varname in ds` this can never be true because the varname should be something like lat_bnds, lon_bnds, rsds etc. So this line always takes the `else` branch - FILLVALUE. 

```suggestion
            source_ds = ds[baseyear]
            ryf[varname].encoding["_FillValue"] = (
                source_ds[varname].encoding.get("_FillValue")
                or FILLVALUE
            )
```

anton-seaice:
Thanks @minghangli-uni - I've made those changes and re-run - still passes repro, see https://github.com/ACCESS-NRI/access-om3-configs/pull/1096#issuecomment-3815606554

minghangli-uni (dismissed):
> Oh I get it now. Um - this is a different scope of work. We can make an new issue for this

Sounds good. in that case, shall we revert this part?

```
# qsub -I -q express -l mem=32GB -l storage=gdata/xp65+gdata/qv56+gdata/tm70+gdata/ua8
```

Otherwise I'm happy to approve this as it stands.

Thanks @anton-seaice

anton-seaice:
Oops - yes, i thought i had done that but it was only in my local copy. SHould be there now!

minghangli-uni (approved):
Looks good to me. Thanks!

PR #113, opened 2026-02-23
Support non-periodic longitude in regional bottom roughness calculations

### LINKED ISSUE #112: bottom roughness: Handle non-periodic longitude case in `fill_missing_data_laplace` for regional configurations
The existing implementation of `fill_missing_data_laplace` only supports periodic longitude that is global configurations. When `periodic_lon_laplace=False`, the function raises a `NotImplementedError`.

I'll implement a non-periodic longitude use case shortly.

minghangli-uni:
closes #112 

In regional grid, longitude does not wrap hence in this PR, only use neighbours that are inside the domain and skip neighbours outside bounds. So this is basically a homogenous Neumann boundary condition


minghangli-uni:
Yeah this is still a draft. There are nans appearing in wet cells for the regional configuration and I am investigating it at the moment.

PR #115, opened 2026-02-26
Use atomic write for bottom_roughness.nc to prevent netcdf corruption

### LINKED ISSUE #114: Netcdf corruption when overwriting bottom_roughness.nc while file is open somewhere
When rerunning the bottom roughness generation workflow and writing to the same output filename `bottom_roughness.nc`, the resulting file can contain extreme float values ~1e308, though the first run producing valid output.

This appears when the file is still open in a Jupyter notebook such as via `xr.open_dataset(...)` while being overwritten. [hdf5 does not handle concurrent open handles and overwrites robustly on shared filesystems.](https://support.hdfgroup.org/documentation/hdf5/latest/_file_lock.html#:~:text=Unfortunately%2C%20this%20file,over%20the%20years.)

My current approach is to write to a temporary file then replace it with the final file. I'll wrap up a PR shortly.

minghangli-uni:
Closes #114 

Now it writes output to a temporary file and atomically replace the final file. So now it prevents file corruption when overwritting existing netcdf files particularly when the file is open in another process, such as notebook.

### DIFF
--- a/external_tidal_generation/generate_bottom_roughness_regrid.py
+++ b/external_tidal_generation/generate_bottom_roughness_regrid.py
@@ -418,8 +418,17 @@ def main():
     global_attrs["inputFile"] = ", ".join(file_hashes)
     regrid_depth_var.attrs.update(global_attrs)
 
-    regrid_depth_var.to_netcdf(args.output_file)
-    print(f"Output written to {args.output_file}")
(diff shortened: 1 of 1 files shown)


aekiss (approved):
LGTM

PR #122, opened 2026-03-30
Add generate_xml_datm_era5.py

ezhilsabareesh8:
This PR adds script to generate `datm.streams.xml` file for ERA5 forcing and addresses issue [548](https://github.com/ACCESS-NRI/access-om3-configs/issues/548)


data_stream_xml_generation/generate_xml_datm_era5.py
```diff
@@ -0,0 +1,221 @@
...
+        if year_first == year_last:
+            file_element = SubElement(datafiles, "file")
+            file_element.text = (
+                f"./INPUT/RYF.{driver_src_var}.{year + 90}_{year + 91}.nc"
+            )
+        else:
+            for month in range(1, 13):  # Loop through months (January to December)
+                days_in_month = calendar.monthrange(year, month)[1]  # Get the number of days in the month
+                file_element = SubElement(datafiles, "file")
+                file_element.text = f"/g/data/rt52/era5/single-levels/reanalysis/{era5_prefix}/{year}/{era5_prefix}_era5_oper_sfc_{year}{month:02d}01-{year}{month:02d}{days_in_month}.nc"
```
anton-seaice:
```suggestion
                file_element.text = f"./INPUT/{era5_prefix}/{year}/{era5_prefix}_era5_oper_sfc_{year}{month:02d}01-{year}{month:02d}{days_in_month}.nc"
```

/g/data/rt52/era5/single-levels/reanalysis

should be listed in `config.yaml` for the configs. Otherwise, the input fixes are not tracked in the payu manifest

Does this add too many files to the manifest ?

anton-seaice: Sorry - we'll need to go one level down, e.g. add:

```
    - /g/data/rt52/era5/single-levels/reanalysis/cp
    - /g/data/rt52/era5/single-levels/reanalysis/lsp
    - /g/data/rt52/era5/single-levels/reanalysis/csf
    - /g/data/rt52/era5/single-levels/reanalysis/lsf
    - /g/data/rt52/era5/single-levels/reanalysis/strd
    - /g/data/rt52/era5/single-levels/reanalysis/ssrd
    - /g/data/rt52/era5/single-levels/reanalysis/ssr
    - /g/data/rt52/era5/single-levels/reanalysis/aluvp
    - /g/data/rt52/era5/single-levels/reanalysis/aluvd
    - /g/data/rt52/era5/single-levels/reanalysis/alnip
    - /g/data/rt52/era5/single-levels/reanalysis/alnid
    - /g/data/rt52/era5/single-levels/reanalysis/sshf
    - /g/data/rt52/era5/single-levels/reanalysis/slhf
  ``` 


 ```suggestion
                file_element.text = f"./INPUT/{year}/{era5_prefix}_era5_oper_sfc_{year}{month:02d}01-{year}{month:02d}{days_in_month}.nc"
```

data_stream_xml_generation/generate_xml_datm_era5.py
```diff
@@ -0,0 +1,221 @@
...
+    if prefix and value.startswith(prefix):
+        return value[len(prefix) :]
+    return value
+
+
+def _git_output(args):
+    return subprocess.check_output(args).decode("ascii").strip()
+
+
+def _compatible_provenance_metadata(file_path, runcmd):
```
anton-seaice: What is this function for ?

What's wrong with https://github.com/ACCESS-NRI/om3-scripts/blob/f0e9985f776deb93a01cf392d192161de8e5e6a4/scripts_common.py#L100-L101 ?

data_stream_xml_generation/generate_xml_datm_era5.py
```diff
@@ -0,0 +1,221 @@
...
+    ),
+    (
+        "ERA5.TDEW",
+        "2d",
+        [("d2m", "Sa_tdew")],
+        "linear",
+    ),
+    ("ERA5.U_10", "10u", [("u10", "Sa_u"), ("u10", "Sa_u10m")], "linear"),
+    ("ERA5.V_10", "10v", [("v10", "Sa_v"), ("v10", "Sa_v10m")], "linear"),
+]
```
anton-seaice, line 70: Could you add a note for which fields are hourly averages, and which are instantaneous ?

It looks like some might be averages, with the with the datestamp at the start of the hourly interval they apply for ? We need to make sure we apply the right time offset for those

data_stream_xml_generation/generate_xml_datm_era5.py
```diff
@@ -0,0 +1,221 @@
...
+
+# Generate stream info elements with changing years
+for stream_name, era5_prefix, datavar_pairs, tintalgo in STREAM_SPECS:
+    stream_info = SubElement(root, "stream_info", name=stream_name)
+    if year_first == year_last:
+        SubElement(stream_info, "taxmode").text = "cycle"
+    else:
+        SubElement(stream_info, "taxmode").text = "extend"
+    SubElement(stream_info, "readmode").text = "single"
+    SubElement(stream_info, "mapalgo").text = "bilinear"
```
anton-seaice, line 110: In https://github.com/ACCESS-NRI/access-om3-configs/issues/243, we're planning on using conservation remapping for fluxes and patch for everything else. Do you wan't to change that now? Or do it is as part of that change ?

data_stream_xml_generation/generate_xml_datm_era5.py
```diff
@@ -0,0 +1,221 @@
...
+SubElement(metadata, "date_generated").text = get_date_generated()
+SubElement(metadata, "history").text = metadata_info
+
+# Generate stream info elements with changing years
+for stream_name, era5_prefix, datavar_pairs, tintalgo in STREAM_SPECS:
+    stream_info = SubElement(root, "stream_info", name=stream_name)
+    if year_first == year_last:
+        SubElement(stream_info, "taxmode").text = "cycle"
+    else:
+        SubElement(stream_info, "taxmode").text = "extend"
```
anton-seaice, line 108: Why did we settle on extend again ? Is this so we can run from the start of 1940 ?

ezhilsabareesh8: Because of partial day start, so we shifted to "extend"

data_stream_xml_generation/generate_xml_datm_era5.py
ezhilsabareesh8: I tried this once while setting up ERA5, but it didn’t work, so I’ll leave this change for later.

data_stream_xml_generation/generate_xml_datm_era5.py
ezhilsabareesh8: Why do we need to go one level deeper here? This is not how it is handled in JRA55-do IAF, where we do not add the variable-specific subdirectory

anton-seaice: There's too many files in `/g/data/rt52/era5/single-levels/reanalysis` 

`payu setup` never finishes

data_stream_xml_generation/generate_xml_datm_era5.py
ezhilsabareesh8: For ERA5, I don’t think we should add a fixed offset 

The instantaneous/state fields are:
- `msl`
- `t2m`
- `d2m`
- `u10`
- `v10`
- `aluvp`
- `aluvd`
- `alnip`
- `alnid`

The interval/accumulation/flux fields are:
- `cp`
- `lsp`
- `csf`
- `lsf`
- `strd`
- `ssrd`
- `ssr`
- `sshf`
- `slhf`

If we were reinterpreting those interval fields as applying to the previous hour, the natural offset would be `-3600`, not the JRA-style `-5400`.

But the file timestamps are not showing a clean, uniform offset pattern that we can safely encode in the generator:
- 1940 instantaneous files start at `00:00`
- 1940 flux/accumulation files start at `07:00`
- recent flux files (for example 2025) also start at `00:00`

So this does not look like the JRA case where one class of files is consistently stamped at `01:30`. It looks more like an early ERA5 coverage difference in some 1940 flux files. A fixed generator offset would misalign later years, and it would not recover the missing first 7 hours in those early files anyway.

So we should keep `offset = 0` for all ERA5 streams unless we decide to make a separate, explicit change for a specific subset/version of the dataset.

data_stream_xml_generation/generate_xml_datm_era5.py
ezhilsabareesh8: Resolved in this commit 32c43a3

data_stream_xml_generation/generate_xml_datm_era5.py
ezhilsabareesh8: Fixed in this commit [5b9d180](https://github.com/ACCESS-NRI/om3-scripts/pull/122/commits/5b9d1804d19d46e9d7f51f66546aad18f2e4d914)

data_stream_xml_generation/generate_xml_datm_era5.py
anton-seaice: > It looks more like an early ERA5 coverage difference in some 1940 flux files.

But for both the older and more recent files, 07:00 means the average of the flux between 07:00 and 08:00 right ? So the middle of the time interval is 07:30. As the timestamp should be in the middle of the time interval they are averaged over, they need an offset of half the interval. (1800s)

ezhilsabareesh8: Good point on interval, but I think the direction is flipped here.

For ERA5 reanalysis, the hourly accumulation / mean-flux fields are defined over the hour ending at the validity time ([refer here under Mean rates/fluxes and accumulations section](https://confluence.ecmwf.int/pages/viewpage.action?pageId=239349050)), not the hour starting at the validity time. So a field stamped `07:00` corresponds to `06:00–07:00`, which means the centered timestamp is `06:30`, not `07:30`.

So if we want to center these fields for linear interpolation, the offset should be `-1800` s, not `+1800` s.

That applies to the hourly interval fields in the ERA5:

`cp`, `lsp`, `csf`, `lsf`, `strd`, `ssrd`, `ssr`, `sshf`, `slhf`

and not to the instantaneous fields:

`msl`, `t2m`, `d2m`, `u10`, `v10`, `aluvp`, `aluvd`, `alnip`, `alnid`

One separate caveat is shortwave when using `coszen` interpolation, in that case the timestamp should be aligned to the start of the interval rather than the midpoint ([refer here](https://escomp.github.io/CTSM/users_guide/setting-up-and-running-a-case/customizing-the-datm-namelist.html#:~:text=For%20coszen%20the%20time%2Dstamps%20of%20the%20data%20should%20correspond%20to%20the%20beginning%20of%20the%20interval%20the%20data%20is%20measured%20for.%20Either%20make%20sure%20the%20time%2Dstamps%20on%20the%20datafiles%20is%20set%20this%20way%2C%20or%20use%20the%20offset%20described%20above%20to%20set%20it.)), so the right offset there would be `-3600` s instead of `-1800` s.

anton-seaice: Sounds good - good sleuthing. Youll make the change ?

I'm ok with switching coszen to linear if you prefer, coszen doesn't work very well - https://github.com/ACCESS-NRI/access-om3-configs/issues/1237

anton-seaice: Make a comment in the file with that link and why the offset is set please too

ezhilsabareesh8: Yep sounds good, will make the change, also switch `coszen` to `linear`

ezhilsabareesh8: Done in this commit 75c4dad

anton-seaice (commented):
Sorry for the drip feed review.

Lets rename

`generate_xml_datm.py` to `generatre_xml_datm_jra55.py`

and

`generate_xml_drof.py` to `generatre_xml_drof_jra55.py`

I think we should change the JRA configs too, to use

/g/data/qv56/replicas/input4MIPs/CMIP6Plus/OMIP/MRI/MRI-JRA55-do-1-6-0/landIce/
/g/data/qv56/replicas/input4MIPs/CMIP6Plus/OMIP/MRI/MRI-JRA55-do-1-6-0/land
/g/data/qv56/replicas/input4MIPs/CMIP6Plus/OMIP/MRI/MRI-JRA55-do-1-6-0/atmos

instead of

/g/data/qv56/replicas/input4MIPs/CMIP6Plus/OMIP/MRI/MRI-JRA55-do-1-6-0/

so that we can reduce the size of the manifests/input.yaml file (which makes payu setup faster)

anton-seaice:
> I think we should change the JRA configs too, to use

See https://github.com/ACCESS-NRI/access-om3-configs/issues/1255

anton-seaice (commented):
looks good

- could you fix the formatting CI failure, and provide repro test results (e.g. a draft PR) for a JRA IAF config please

ezhilsabareesh8:
Thanks @anton-seaice , here is the repro test 
https://github.com/ACCESS-NRI/access-om3-configs/pull/1269#issuecomment-4241708427

anton-seaice (approved):
Thanks @ezhilsabareesh8

PR #124, opened 2026-04-28
regrid_forcing.py: Add option to homogenize after regridding

dougiesquire:
This PR adds a `--homogenize` option to `regrid_forcing.py`. The option replaces regridded values with their horizontal mean after regridding. This can be useful for idealised/test configurations, e.g. [this single-column configuration](https://github.com/dougiesquire/om3-mom6-single-column/tree/jra_ryf/EPBL) which uses the MOM6 option `HOMOGENIZE_FORCINGS = True`.

This does not change answers by default and I've checked that it does what I want when turned on (example output here: `/g/data/tm70/ds0092/model/config/om3_M_single-column/EPBL/INPUT/BATS/SFe_Hamiltonetal2020_monthly_clim.nc`)


anton-seaice (approved):
Looks good

PR #131, opened 2026-06-15
era5_rechunking: add script to rechunk and concatenate ERA5 forcing files

ezhilsabareesh8:
Adds `era5_rechunking/` with a Python script and PBS script to rechunk ERA5 single-level monthly files from the `rt52` archive into yearly files optimised for ACCESS-OM3 / DATM.

Related issue https://github.com/ACCESS-NRI/access-om3-configs/issues/1381

What it does:

- Reads directly from /g/data/rt52/era5/single-levels/reanalysis/{stream}/{year}/
- Rechunks from [93, 91, 180] (ERA5 default) → [1, 721, 1440] (one full-resolution timestep per chunk)
- Concatenates 12 monthly files into one yearly file per stream
- Preserves raw int16 values bit-for-bit (no unpack/repack)
- Covers all 13 ERA5 streams, 1940–2026

Test output is here `/g/data/tm70/ek4684/era5_rechunk_test/`


anton-seaice (commented):
This appears somewhat complicated 

- are using xarray (with Dask) or nco/ncks not good options ?
- do we need this level of configuration in input options ?

anton-seaice:
I was thinking something like this:

```
import xarray as xr

from dask.distributed import Client

client = Client()

client.dashboard_link

ds = xr.open_mfdataset(
    '/g/data/rt52/era5/single-levels/reanalysis/2t/1940/*.nc',
    chunks={'latitude':-1,'longitude':-1}
)

ds_new = ds.t2m.chunk({'time':1})

ds_new.to_netcdf('2t_1940_era5.nc')
```

is that too slow ?

ezhilsabareesh8:
> is that too slow ?

Changed to xarray in this commit [4b0efbf](https://github.com/ACCESS-NRI/om3-scripts/pull/131/commits/4b0efbf77a4926941a9b9084a2f94c0f2d710fc1), it is taking 3 minutes per variable now

era5_rechunking/run_make_era5_yearly_rechunked.sh
```diff
@@ -0,0 +1,31 @@
...
+#PBS -N era5_yearly_rechunk
+#PBS -P tm70
+#PBS -q normal
+#PBS -l ncpus=24
+#PBS -l mem=96GB
+#PBS -l walltime=24:00:00
+#PBS -l storage=gdata/tm70+gdata/xp65+gdata/rt52
+#PBS -l wd
+#PBS -j oe
+#PBS -o /g/data/tm70/ek4684/era5_rechunked_1h/logs/yearly/
```
anton-seaice: for sharability, probably best not to set this
```suggestion
```

era5_rechunking/run_make_era5_yearly_rechunked.sh
```diff
@@ -0,0 +1,31 @@
...
+
+set -euo pipefail
+
+SCRIPT_DIR="${PBS_O_WORKDIR}"
+
+# ── user configuration ─────────────────────────────────────────────────────────
+# Directory where yearly rechunked files will be written:
+#   {OUTPUT_DIR}/{stream}/{stream}_era5_oper_sfc_{YYYYMMDD}-{YYYYMMDD}.nc
+OUTPUT_DIR="/g/data/tm70/ek4684/era5_rechunked_1h_yearly"
```
anton-seaice:
```suggestion
OUTPUT_DIR="/g/data/${PROJECT}/${USER}/era5_rechunked_1h_yearly"
```

or similar might me more shareable

era5_rechunking/run_make_era5_yearly_rechunked.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/bash
+# Copyright 2025 ACCESS-NRI and contributors. See the top-level COPYRIGHT file for details.
```
anton-seaice:
```suggestion
# Copyright 2026 ACCESS-NRI and contributors. See the top-level COPYRIGHT file for details.
```

era5_rechunking/run_make_era5_yearly_rechunked.sh
anton-seaice, line 1: is `submit_era5_yearly_rechuncked.sh` a clearer filename ? It's consistent with https://github.com/ACCESS-NRI/om3-scripts/blob/main/external_tidal_generation/submit_bottom_roughness.sh then ?

era5_rechunking/run_make_era5_yearly_rechunked.sh
anton-seaice: I think put it on /scratch rather than /g/data/ ?

era5_rechunking/run_make_era5_yearly_rechunked.sh
```diff
@@ -0,0 +1,31 @@
...
+# ── user configuration ─────────────────────────────────────────────────────────
+# Directory where yearly rechunked files will be written:
+#   {OUTPUT_DIR}/{stream}/{stream}_era5_oper_sfc_{YYYYMMDD}-{YYYYMMDD}.nc
+OUTPUT_DIR="/g/data/tm70/ek4684/era5_rechunked_1h_yearly"
+# ──────────────────────────────────────────────────────────────────────────────
+
+module use /g/data/xp65/public/modules
+module load conda/analysis3-26.02
+
+python3 "${SCRIPT_DIR}/make_era5_yearly_rechunked.py" \
```
anton-seaice: SCRIPT_DIR is only used in one place, so may as just use PBS_O_WORKDIR directly here

era5_rechunking/make_era5_yearly_rechunked.py
```diff
@@ -0,0 +1,372 @@
+#!/usr/bin/env python3
+# Copyright 2025 ACCESS-NRI and contributors. See the top-level COPYRIGHT file for details.
```
anton-seaice:
```suggestion
# Copyright 2026 ACCESS-NRI and contributors. See the top-level COPYRIGHT file for details.
```

era5_rechunking/make_era5_yearly_rechunked.py
```diff
@@ -0,0 +1,372 @@
...
+    source_files = sorted(source_dir.glob("*.nc"))
+    if not source_files:
+        logging.error(f"No .nc files in {source_dir}")
+        return False
+
+    # Derive output filename from first/last source file date stamps
+    m0 = re.search(r"(\d{8})-(\d{8})", source_files[0].name)
+    ml = re.search(r"(\d{8})-(\d{8})", source_files[-1].name)
+    start = m0.group(1) if m0 else f"{year}0101"
+    end = ml.group(2) if ml else f"{year}1231"
```
anton-seaice:
```
    end = ml.group(2) if ml else f"{year}1231"
```

In what case does this else statement get reached ?

era5_rechunking/make_era5_yearly_rechunked.py
```diff
@@ -0,0 +1,372 @@
...
+    if not source_dir.is_dir():
+        logging.error(f"Source directory not found: {source_dir}")
+        return False
+
+    source_files = sorted(source_dir.glob("*.nc"))
+    if not source_files:
+        logging.error(f"No .nc files in {source_dir}")
+        return False
+
+    # Derive output filename from first/last source file date stamps
```
anton-seaice:
```suggestion
    # In case this year is a partially year, find first/last source file date stamps
```

era5_rechunking/make_era5_yearly_rechunked.py
anton-seaice: Why are we getting dates from the filenames rather than the data ?

e.g. would

```
start = str(ds.time.min().dt.strftime("%Y%m%d").values)
end = str(ds.time.max().dt.strftime("%Y%m%d").values)
```

be better ?

anton-seaice (commented):
ill get back to this tomorrow

era5_rechunking/make_era5_yearly_rechunked.py
```diff
@@ -0,0 +1,372 @@
...
+        default=1,
+        help="Number of parallel workers for full run (default: 1).",
+    )
+    parser.add_argument("--source-dir", default=SOURCE_BASE)
+    parser.add_argument("--output-dir", default=OUTPUT_BASE)
+    parser.add_argument("--overwrite", action="store_true")
+    args = parser.parse_args()
+
+    # ── single-task mode ──────────────────────────────────────────────────────
+    if args.stream and args.year:
```
anton-seaice: Do we need this section?

Can args.stream be a list of streams, and args.year be a range of years to process? And then just treat them the same as "multi-task mode"

era5_rechunking/make_era5_yearly_rechunked.py
```diff
@@ -0,0 +1,372 @@
...
+        return
+
+    t_start = time.time()
+    failures = []
+
+    if args.workers == 1:
+        for task in tasks:
+            stream, year, rc, _ = _subprocess_task(task)
+            if rc != 0:
+                failures.append((stream, year))
```
anton-seaice:
```    if args.workers == 1:
        for task in tasks:
            stream, year, rc, _ = _subprocess_task(task)
            if rc != 0:
                failures.append((stream, year))
```

Do we need this?

It appears to work fine to comment this out and use `multiprocessing.Pool(args.workers)` with 1 worker

era5_rechunking/make_era5_yearly_rechunked.py
```diff
@@ -0,0 +1,372 @@
...
+    "cp": "cp",
+    "csf": "csf",
+    "lsf": "lsf",
+    "lsp": "lsp",
+    "msl": "msl",
+    "ssr": "ssr",
+    "ssrd": "ssrd",
+    "strd": "strd",
+}
+
```
anton-seaice:
```suggestion
# Streams used by ERA5 datamode in OM3
# Stream : variable name mapping in ERA5 data
STREAM_TO_VARNAME = {
    "10u": "u10",
    "10v": "v10",
    "2d": "d2m",
    "2t": "t2m",
    "ci": "siconc",
    "cp": "cp",
    "csf": "csf",
    "lsf": "lsf",
    "lsp": "lsp",
    "msl": "msl",
    "ssr": "ssr",
    "ssrd": "ssrd",
    "strd": "strd",
}

ALL_STREAMS = list(STREAM_TO_VARNAME.keys())

```

era5_rechunking/make_era5_yearly_rechunked.py
```diff
@@ -0,0 +1,372 @@
...
+    "ssrd",
+    "strd",
+]
+
+STREAM_TO_VARNAME = {
+    "10u": "u10",
+    "10v": "v10",
+    "2d": "d2m",
+    "2t": "t2m",
+    "ci": "siconc",
```
anton-seaice: why is siconc in this list ?

era5_rechunking/make_era5_yearly_rechunked.py
```diff
@@ -0,0 +1,372 @@
...
+        for year in range(YEAR_FIRST, YEAR_LAST + 1):
+            if _output_exists(stream, year, args.output_dir) and not args.overwrite:
+                skipped += 1
+            else:
+                tasks.append(
+                    (stream, year, args.source_dir, args.output_dir, args.overwrite)
+                )
+
+    print(f"Tasks to run:  {len(tasks)}")
+    print(f"Tasks skipped: {skipped}  (output already exists)")
```
anton-seaice:
```suggestion
    if not args.overwrite:
        print(f"Tasks skipped: {skipped}  (output already exists)")
```

era5_rechunking/make_era5_yearly_rechunked.py
```diff
@@ -0,0 +1,372 @@
...
+        cmd.append("--overwrite")
+
+    t0 = time.time()
+    result = subprocess.run(cmd, capture_output=True, text=True)
+    elapsed = time.time() - t0
+
+    status = "OK" if result.returncode == 0 else "FAIL"
+    if result.returncode != 0:
+        tail = (result.stderr or result.stdout or "").strip()[-500:]
+        print(f"[{status}] {stream}/{year} ({elapsed:.0f}s)\n  {tail}", flush=True)
```
anton-seaice: When it fails it didn't give enough useful information to understand why:

This is the log extract

```
[FAIL] 10u/1954 (161s)
  /g/data/tm70/as2285/making_inputs/om3-scripts/scripts_common.py:118: UserWarning: /g/data/tm70/as2285/making_inputs/om3-scripts/era5_rechunking/make_era5_yearly_rechunked.py contains uncommitted changes! Commit and push your changes before generating any production output.
  warn(
```

era5_rechunking/run_make_era5_yearly_rechunked.sh
ezhilsabareesh8: Yes, agreed. I renamed it to `submit_era5_yearly_rechunked.sh` for consistency with the other submit scripts

era5_rechunking/make_era5_yearly_rechunked.py
```diff
@@ -0,0 +1,372 @@
...
+    # without this xarray unpacks to float64 (8x storage, wrong output dtype).
+    # Full-spatial Dask chunks (one per monthly file) avoids the 8x8=64 source-chunk
+    # reads that would otherwise be needed per output timestep when the source spatial
+    # chunks (91x180) differ from the output (721x1440). The encoding dict controls
+    # the on-disk HDF5 chunk layout independently of the Dask chunk size.
+    ds = xr.open_mfdataset(
+        [str(f) for f in source_files],
+        combine="by_coords",
+        mask_and_scale=False,
+        chunks={"time": 744, "latitude": 721, "longitude": 1440},
```
anton-seaice: There isn't enough RAM in the qsub job to do this operation - it would need >10GB per task. by the time each whole file is uncompressed.

era5_rechunking/run_make_era5_yearly_rechunked.sh
ezhilsabareesh8: I changed the default submit-script output path to `/scratch/${PROJECT}/${USER}/era5_rechunked_1h_yearly`

era5_rechunking/run_make_era5_yearly_rechunked.sh
ezhilsabareesh8: Removed SCRIPT_DIR and used ${PBS_O_WORKDIR} directly

era5_rechunking/make_era5_yearly_rechunked.py
```diff
@@ -0,0 +1,372 @@
...
+    )
+
+    # Provenance
+    this_file = os.path.normpath(__file__)
+    now_iso = datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ")
```
anton-seaice:
```suggestion
    date_str = datetime.datetime.now().strftime("%Y-%m-%dT%H:%M:%SZ")
```
we don't normally use utc time for stuff

era5_rechunking/make_era5_yearly_rechunked.py
ezhilsabareesh8: Yes, using the data is better. I changed this to derive start and end from `ds.time.min()` and `ds.time.max()`, so partial-year files reflect their actual time coverage instead of relying on filename parsing

ezhilsabareesh8: That fallback was defensive for filenames that did not match the expected `YYYYMMDD-YYYYMMDD` pattern, but it is no longer needed because the output dates now come from the dataset time coordinate.

era5_rechunking/make_era5_yearly_rechunked.py
anton-seaice: Not really sure what to suggest

era5_rechunking/make_era5_yearly_rechunked.py
ezhilsabareesh8: Changed in this commit [4086207](https://github.com/ACCESS-NRI/om3-scripts/pull/131/commits/408620701f7f78d43b36c882dfe45f9ae11667c6)

ezhilsabareesh8:
@anton-seaice Test output of the latest commit is here `/scratch/tm70/ek4684/era5_rechunk_test/csf/csf_era5_oper_sfc_19790101-19791231.nc `

era5_rechunking/make_era5_yearly_rechunked.py
```diff
@@ -0,0 +1,372 @@
...
+        last_line = (result.stdout.strip().split("\n") or [""])[-1]
+        print(f"[{status}] {stream}/{year} ({elapsed:.0f}s)  {last_line}", flush=True)
+
+    return stream, year, result.returncode, elapsed
+
+
+def _output_exists(stream, year, output_base):
+    return bool(
+        glob.glob(
+            os.path.join(output_base, stream, f"{stream}_era5_oper_sfc_{year}*-*.nc")
```
anton-seaice: This method a bit dangerous because it doesn't detect if the file is complete or not, but i dont have a better suggestion

anton-seaice (commented):
This didn't work for me (using the submission script).

Requesting one node on sapphirerapids and then only running 12 workers at once was ok - but i don't really undersand why it needs that much RAM to succeed

anton-seaice (changes requested):
I started looking at 10u, the 1940 january data matches, but later months don't appear to match

Image

.

ezhilsabareesh8:
> I started looking at 10u, the 1940 january data matches, but later months don't appear to match

Thanks @anton-seaice for flagging this. The mismatch came from the monthly ERA5 files using different `scale_factor` and `add_offset` values.

I’ve updated the script so it now decodes each monthly file with its own packing metadata, then repacks into a single yearly `int16` encoding:

https://github.com/ACCESS-NRI/om3-scripts/pull/131/commits/19686b9c16bb315a5db5221f86acba68af53b923

So the raw packed integers are not expected to be bit-for-bit identical when monthly scale/offset values differ. The validation now checks decoded physical values within half of the yearly packing step, with missing-value masks preserved exactly.

I tested the regenerated `csf/1979` output here and it passes the monthly validation.

```text
/scratch/tm70/ek4684/era5_rechunk_test/csf/
```

For the regenerated `10u/1940` output, the maximum absolute decoded difference is `0.000686379 m s**-1`, which is within the expected repacking tolerance of `0.000686380 m s**-1` (half of the yearly `scale_factor`). This is consistent with normal int16 quantisation after decoding monthly files and repacking them into a single yearly encoding.

compare_10u_monthly_rechunked

ezhilsabareesh8:
<html><head></head><body><p>Thanks @dougiesquire, I tested your suggestion by writing the decoded monthly ERA5 values into a yearly unpacked <code inline="">float64</code> file, without <code inline="">scale_factor</code> or <code inline="">add_offset</code>.</p><p>Float64 test output:</p><pre><code class="language-text">/scratch/tm70/ek4684/era5_rechunk_test_float/float64/10u/10u_era5_oper_sfc_19400101-19401231_float64.nc
</code></pre><p>Packed <code inline="">int16</code> comparison output:</p><pre><code class="language-text">/scratch/tm70/ek4684/tmp/era5_rechunk_single_test/10u/10u_era5_oper_sfc_19400101-19401231.nc
</code></pre><p>For <code inline="">10u/1940</code>, I compared the decoded monthly source fields against the yearly <code inline="">float64</code> output using the first timestep from every month. The decoded values are preserved exactly in the <code inline="">float64</code> case.</p>

output | exact decoded-value equality | max abs difference | size | relative size
-- | -- | -- | -- | --
packed int16 | no, repacked | within the expected rounding error from repacking | 11.207 GiB | 1.00x
unpacked float64 | yes | 0 | 55.538 GiB | 4.96x

<p>So the <code inline="">float64</code> option does preserve the decoded monthly values exactly, but the file is about 5x larger for this variable/year.</p><p>Given that, I think the practical options are either to keep packed <code inline="">int16</code> and document the expected repacking tolerance, or avoid monthly-to-yearly concatenation and keep the original one-file-per-month structure.</p></body></html>

compare_10u_monthly_float64

anton-seaice (commented):
I was able to generate two years out data without trouble using these scripts

I then reviewed output only:

We should probably remove this part from the `summary` in attribtues:

> This file is part of the ERA5 replica hosted at NCI Australia. For more information please see http://dx.doi.org/10.25914/5f48874388857

Also this line is hardcoded but could be ignored;

> rechunked_by :
>     Ezhilsabareesh Kannadasan (ek4684)
> 

for `u10`, `v10` for year 2000:

numpy.allclose is true with

`atol = 1e-3, rtol = 1e-6, equal_nan = True`

The absolute tolerance needs to be fairly loose (as `integer*scale + offset` precision changes depending on the magnitude of the value).

for `msl` in year 2000:

`atol = 1e-1, rtol = 1e-6, equal_nan = True` is True everywhere

ezhilsabareesh8:
Thanks @anton, this should be addressed now in this commit a9f68a1.

I removed the NCI ERA5 replica sentence from the copied `summary` attribute.

For the tolerance checks, I have not hardcoded the `u10/v10` or `msl` tolerances. Instead, the script now derives the decoded-value tolerance from each output file’s own yearly `scale_factor` using `0.5 * abs(scale_factor)`, and stores this in the variable metadata as `repacking_absolute_tolerance`.

The validation now uses this same per-file tolerance when comparing decoded source and output values. The 2-year all-variable test output is here `/scratch/tm70/ek4684/era5_rechunk_2yr_allvars_test/`

anton-seaice (approved):
Thanks @ezhilsabareesh8

PR #133, opened 2026-06-16
Replace standardise_mom6_filenames.sh with a new script

### LINKED ISSUE #132: standardise_mom6_filenames.sh doesn't support non-annual filenames
The script `standardise_mom6_filenames.sh` only removes underscores before the first instance of four consecutive digits. Aside from being somewhat unrobust, this also doesn't support non-annual files, e.g. `*_2024_01.nc`.

We could also extend the script to support OM2 output so the same script can be used in both places

dougiesquire:
The new script:

- Supports more than just annual timestamps (e.g. _2023_01_01)
- Supports OM2 output as well

Closes #132


dougiesquire:
I've tested that this:

- Works as expected as a archive userscript with ACCESS-OM2 output (yearly and monthly output)
- Doesn't change things for ACCESS-OM3 output

payu_config/archive_scripts/standardise_mom_filenames.py
```diff
@@ -0,0 +1,95 @@
...
+    For example:
+        file._2023.nc       -> file.2023.nc
+        file._2023_01.nc    -> file.2023-01.nc
+        file._2023_01_15.nc -> file.2023-01-15.nc
+    """
+    dirpath = os.path.dirname(path)
+    basename = os.path.basename(path)
+
+    # Split the .nc extension
+    ext_match = re.match(r"^(.*?)(\.nc.*)$", basename)
```
anton-seaice, line 33: I think the matches against things where there is anything(`.*`) after `.nc`, is that needed ?

Probably harmless even if not needed

payu_config/archive_scripts/standardise_mom_filenames.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+    args = parser.parse_args()
+
+    if args.out_dir:
+        if not os.path.isdir(args.out_dir):
+            print(f"{args.out_dir} does not exist")
+            sys.exit(1)
+        out_dirs = [args.out_dir]
+    else:
+        out_dirs = sorted(glob.glob("archive/output*[0-9]"), reverse=True)
```
anton-seaice: This matches lots of things that aren't the expcted pattern:

archive/output1
archive/output123
archive/output_2024_01_01_0
archive/output-test9

probably harmless ?

anton-seaice: I see you could argue this is no worse than the old method

test/test_payu_conf/test_mom_filenames.py
```diff
@@ -0,0 +1,149 @@
...
+    output_dir = tmp_path / hist_dir
+    file_dir = output_dir / file_subdir if file_subdir else output_dir
+
+    original_paths = make_files(str(file_dir), diag_base, suffixes)
+    chdir(tmp_path)
+
+    # Create files at the expected destination paths so there is something to protect
+    expected_paths = [standardised_path(p, s) for p, s in zip(original_paths, suffixes)]
+    for p in expected_paths:
+        Path(p).touch()
```
anton-seaice, line 142: these files should probably be made in a fixture, otherwise they might impact later tests right ?

anyway, again you can say "no worse than the old method"

anton-seaice (dismissed):
Code review only - looks good to me

payu_config/archive_scripts/standardise_mom_filenames.py
dougiesquire: Yeah, we also want to match uncollated files e.g. `*.nc.0001`

payu_config/archive_scripts/standardise_mom_filenames.py
dougiesquire: Yeah, I just copied this from the old approach. I think the intended behaviour was any number of digits so I'll change to do that

dougiesquire: Actually, I've changed my mind. I'll change to `"archive/output*`

test/test_payu_conf/test_mom_filenames.py
dougiesquire: This is using the pytest `tmp_path` fixture so I think it is safe. Files created in one tests `tmp_path` are invisible to others.

I think the `chdir` could be an issue if we weren't running this sequentially, but we are.

payu_config/archive_scripts/standardise_mom_filenames.py
dougiesquire:
```suggestion
        out_dirs = sorted(glob.glob("archive/output*"), reverse=True)
```

dougiesquire: Done in [4874a0d](https://github.com/ACCESS-NRI/om3-scripts/pull/133/commits/4874a0d346f3fc8977e2046da8b63a7151fc078f)

anton-seaice (approved):

PR #137, opened 2026-06-26
Fix up target_cells for rof weights generation

### LINKED ISSUE #136: ROF remapping weights generation very slow
There is an error in the ROF remap weights generation, which causes it to be very slow.

In these lines:

https://github.com/ACCESS-NRI/om3-scripts/blob/0e7124917f2fe7149ec9324d87c0041e2dfa6d30/mesh_generation/generate_rof_weights.py#L101-L106

`target_cells` generates a landmask for cells where runoff can go to, and then the index for these cells are found by multiplying the mask array by the indexes. All land cells then end up being index 0 in `mask_target`, giving a very large array with mostly zeros. 

This then makes the resulting BallTree very slow to query.

anton-seaice:
Closes #136 

This change removes [0,0] (the bottom left corner of the grid) from the possible cells runoff can be mapped to. This also reduces the size of the target_cells array by removing all the previous 0,0 entries.

This is answer changing because no runoff is mapped to 0,0, and for whatever reason, some runoff appears to get mapped to slightly different cells (which are at very similar distance from to the old cell it was mapped to)

See [notebook](https://github.com/anton-seaice/sandbox/blob/d96f985163038b05441e945f609d37bb41a994a3/om3_remapping/om3-scripts-136-faster_weights.ipynb)

This is the diff in remapping unity input everywhere between new_weights - old_weights
image

Note 
- some islands and peninsulas have water remapped to the other side of them (but not in a consistent direction between methods). 
- water from the Antarctic continent is mapped to the coast now (rather than bottom left corner)

### DIFF
--- a/mesh_generation/generate_rof_weights.py
+++ b/mesh_generation/generate_rof_weights.py
@@ -112,7 +112,7 @@ def drof_remapping_weights(mesh_filename, weights_filename, nx, ny, global_attrs
         center_coords_rad.where(target_cells_da, drop=True), metric="haversine"
     )
 
-    # Index for the target_cells
+    # Index for the target_cells (other=0 preserves integer dtype)
     target_cells_i = mod_mesh_ds.elementCount.where(target_cells_da, other=0, drop=True)
 
     # Using the Tree, look up the nearest ocean cell to every destination grid cell in our weights file. Note our
--- a/mesh_generation/generate_rof_weights.py
+++ b/mesh_generation/generate_rof_weights.py
@@ -99,18 +99,18 @@ def drof_remapping_weights(mesh_filename, weights_filename, nx, ny, global_attrs
     land_neighbours = binary_dilation(mask_2d == 0)
 
     # target for runoff is ocean cells which are adjacent land
-    target_cells = (land_neighbours & mask_2d).flatten()
-
-    # Find index for all target cells
-    mask_target = (mod_mesh_ds.elementCount * target_cells).astype("int")
+    target_cells = ((land_neighbours & mask_2d) == 1).flatten()
 
     # Haversine distances expect lat first, lon second, so index coordDim backwards
     center_coords_rad = deg2rad(mod_mesh_ds.centerCoords.isel(coordDim=[1, 0]))
 
     # Make a BallTree from the ocean cells
     mask_tree = BallTree(
-        center_coords_rad.isel(elementCount=mask_target), metric="haversine"
+        center_coords_rad.isel(elementCount=target_cells), metric="haversine"
     )
+
+    # Index for the target_cells
+    target_cells_i = mod_mesh_ds.elementCount.sel(elementCount=target_cells)
 
     # Using the Tree, look up the nearest ocean cell to every destination grid cell in our weights file. Note our
     # weights are indexed from 1 (i.e. Fortran style) but xarray starts from 0 (i.e. python style), so subract one from
@@ -120,7 +120,7 @@ def drof_remapping_weights(mesh_filename, weights_filename, nx, ny, global_attrs
         center_coords_rad.isel(elementCount=(weights_ds.row - 1)), return_distance=False
     )
 
-    new_row = mask_target[ii[:, 0]] + 1
+    new_row = target_cells_i[ii[:, 0]] + 1
 
     # Get the mesh element areas and adjust:
     # n.b. per CMEPS we are using the internally calculated areas, not the user provided ones.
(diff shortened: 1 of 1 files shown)


anton-seaice:
This works now for the panant grid, takes about ~10mins to complete

mesh_generation/generate_rof_weights.py
```diff
@@ -99,19 +99,19 @@ def drof_remapping_weights(mesh_filename, weights_filename, nx, ny, global_attrs
     land_neighbours = binary_dilation(mask_2d == 0)
 
     # target for runoff is ocean cells which are adjacent land
-    target_cells = (land_neighbours & mask_2d).flatten()
-
-    # Find index for all target cells
-    mask_target = (mod_mesh_ds.elementCount * target_cells).astype("int")
+    target_cells = ((land_neighbours & mask_2d) == 1).flatten()
```
anton-seaice, line 102: I confirmed that the old:

`center_coords_rad.isel(elementCount=mask_target).where(target_cells, drop=True) `

is identical to the new:

`center_coords_rad.isel(elementCount=target_cells)`

mesh_generation/generate_rof_weights.py
```diff
@@ -99,19 +99,19 @@ def drof_remapping_weights(mesh_filename, weights_filename, nx, ny, global_attrs
...
 
     # Haversine distances expect lat first, lon second, so index coordDim backwards
     center_coords_rad = deg2rad(mod_mesh_ds.centerCoords.isel(coordDim=[1, 0]))
 
     # Make a BallTree from the ocean cells
     mask_tree = BallTree(
-        center_coords_rad.isel(elementCount=mask_target), metric="haversine"
+        center_coords_rad.isel(elementCount=target_cells), metric="haversine"
```
dougiesquire: Could we use `where` here for clarity?

mesh_generation/generate_rof_weights.py
```diff
@@ -99,19 +99,19 @@ def drof_remapping_weights(mesh_filename, weights_filename, nx, ny, global_attrs
...
     )
 
+    # Index for the target_cells
+    target_cells_i = mod_mesh_ds.elementCount.sel(elementCount=target_cells)
```
dougiesquire: Could we use `where` here for clarity?

dougiesquire (commented):
Thanks @anton-seaice - good catch

I think using `.where(target_cells, drop=True)` would be clearer here. I wasn't actually aware you could pass `isel`/`sel` boolean arrays and I can't find it documented in the xarray docs.

dougiesquire:
> I think using `.where(target_cells, drop=True)` would be clearer here. I wasn't actually aware you could pass `isel`/`sel` boolean arrays and I can't find it documented in the xarray docs.

E.g. something like (untested code)

```python
target_cells_da = xr.DataArray(target_cells, dims='elementCount')

mask_tree = BallTree(
    center_coords_rad.where(target_cells_da, drop=True), metric="haversine"
)

target_cells_i = mod_mesh_ds.elementCount.where(target_cells_da, drop=True)
```

dougiesquire:
> E.g. something like (untested code)
> 
> ```python
> target_cells_da = xr.DataArray(target_cells, dims='elementCount')
> 
> mask_tree = BallTree(
>     center_coords_rad.where(target_cells_da, drop=True), metric="haversine"
> )
> 
> target_cells_i = mod_mesh_ds.elementCount.where(target_cells_da, drop=True)
> ```

I just tested this, and you'll want to provide an `other` argument to `where` to prevent it from converting the ints to floats, e.g.

```python
target_cells = xr.DataArray(
    ((land_neighbours & mask_2d) == 1).flatten(),
    dims='elementCount'
)

mask_tree = BallTree(
    center_coords_rad.where(target_cells, other=0, drop=True), metric="haversine"
)

target_cells_i = mod_mesh_ds.elementCount.where(target_cells, other=0, drop=True)
```

anton-seaice:
Thanks @dougiesquire - i tested there is no change in output when using the updated code

mesh_generation/generate_rof_weights.py
```diff
@@ -99,19 +99,22 @@ def drof_remapping_weights(mesh_filename, weights_filename, nx, ny, global_attrs
...
+    # convert back to a DataArray
+    target_cells_da = xr.DataArray(target_cells, dims="elementCount")
 
     # Haversine distances expect lat first, lon second, so index coordDim backwards
     center_coords_rad = deg2rad(mod_mesh_ds.centerCoords.isel(coordDim=[1, 0]))
 
     # Make a BallTree from the ocean cells
     mask_tree = BallTree(
-        center_coords_rad.isel(elementCount=mask_target), metric="haversine"
+        center_coords_rad.where(target_cells_da, other=0, drop=True), metric="haversine"
```
dougiesquire: Sorry, the `other` is not actually needed here

```suggestion
        center_coords_rad.where(target_cells_da, drop=True), metric="haversine"
```

mesh_generation/generate_rof_weights.py
```diff
@@ -99,19 +99,22 @@ def drof_remapping_weights(mesh_filename, weights_filename, nx, ny, global_attrs
...
     )
 
+    # Index for the target_cells
```
dougiesquire:
```suggestion
    # Index for the target_cells (other=0 preserves integer dtype)
```

dougiesquire (dismissed):

dougiesquire (approved):
Thanks @anton-seaice

Issue #142, opened 2026-07-02
Update drof weights to include spreading


anton-seaice:
I'm planning a three PR approach to [spread large rivers](https://github.com/ACCESS-NRI/access-om3-configs/issues/396#issuecomment-4427325574) using the drof weights

1. Write a test to show that the weights are conservative
2. Remove the dependency on esmf in creating weights, and allow for different input/out meshes
3. Add spreading

anton-seaice:
There is part 4 - support mapping from UM mesh, 

can move mesh code from

https://github.com/ACCESS-NRI/land-fraction-ancillary-suite/blob/7c3f54084673eb0d0513058c4c6282ab74ead3e4/bin/create_landfrac.py#L41

to this repo

PR #145, opened 2026-07-15
Runoff spreading

anton-seaice:
This generates runoff spreading weights as discussed in https://github.com/ACCESS-NRI/access-om3-configs/issues/396

For the 6 large rivers, runoff is spread to nearby ocean cells with exponentially decaying weights.
Everywhere else, runoff if mapped to closest ocean cell only

Contibutes to #142


anton-seaice:
Example output in /g/data/tm70/as2285/making_inputs/om3-scripts/mesh_generation/25km_remap_weights.nc

anton-seaice:
And data in /g/data/tm70/as2285/payu/om3_rofl_spread/MC_25km_jra_iaf_spread/archive

mesh_generation/generate_rof_weights.py
```diff
@@ -62,85 +85,214 @@ def mesh_area(mesh_filename):
...
-    row = target_cells_i[ii[:, 0]]
+            new_area = area[row]
 
-    # Get the mesh element areas and adjust for area change between the
-    # source cell (col) and destination (row)
-    area = mesh_area(mesh_filename)
-    old_area = area[col]
-    new_area = area[row]
+            # Exponential weights
+            w = np.exp(-dist_N[spread_i] / FOLD)
```
ezhilsabareesh8: @anton-seaice I think there is a units issue in the decay calculation here:

`w = np.exp(-dist_N[spread_i] / FOLD)`

Because the tree uses metric="haversine", dist_N is returned as an angular distance in radians. It is then divided directly by `FOLD = 15`, with no conversion to kilometres.

I checked this against the generated 25 km weights file as well. The 75 destination cells are roughly 125–345 km away, corresponding to about 0.020–0.054 radians. Using the current calculation gives decay factors of around 0.997–0.999, which matches the almost-flat weights in the file. The weights therefore vary by only about 0.2% and the runoff is being spread nearly uniformly rather than noticeably decaying with distance.

Could we convert the haversine distances to a physical unit before applying the decay, and make the unit of FOLD explicit?

mesh_generation/generate_rof_weights.py
```diff
@@ -62,85 +85,214 @@ def mesh_area(mesh_filename):
...
+        )
+
+        if self.spread:
+            # nearest N - spread
+            dist_N, i_N = mask_tree_spread.query(
+                in_coords_rad,
+                k=SPREAD_N,
+                return_distance=True,
+            )
+
```
ezhilsabareesh8, line 198: Could we guard against meshes with fewer than SPREAD_N eligible ocean cells here?
```
dist_N, i_N = mask_tree_spread.query(
    in_coords_rad,
    k=SPREAD_N,
    return_distance=True,
)
```

Perhaps use min(SPREAD_N, number_of_target_cells) and carry that effective value through the downstream array allocations and loops.

mesh_generation/generate_rof_weights.py
```diff
@@ -62,85 +85,214 @@ def mesh_area(mesh_filename):
...
+        row += 1
+        col += 1
+
+        weights_ds = xr.Dataset()
+        weights_ds["col"] = xr.DataArray(data=col, dims="n_s")
+        weights_ds["row"] = xr.DataArray(data=row, dims="n_s")
+        weights_ds["S"] = xr.DataArray(data=S, dims="n_s")
+
+        # Remove zero weights
+        weights_ds = weights_ds.where(weights_ds["S"] != 0, drop=True)
```
ezhilsabareesh8: This line upcasts col and row from integers to float64, because where() temporarily inserts NaN values. The generated weights file confirms both variables are stored as floats.

Since col and row are ESMF indices, could we keep or cast them back to integer after filtering?

ezhilsabareesh8 (changes requested):

mesh_generation/generate_rof_weights.py
anton-seaice: Thanks! Will fix

mesh_generation/generate_rof_weights.py
anton-seaice: I think BallTree returns an error, which is fine, Very small meshes with runoff in them aren't very plausible

mesh_generation/generate_rof_weights.py
anton-seaice: Thank - fixed by adding `other = 0`

added some tests for this too

anton-seaice:
Ready to go @ezhilsabareesh8

mesh_generation/generate_rof_weights.py
```diff
@@ -60,8 +60,9 @@
 # Number of cells to spread to
 SPREAD_N = 75
 
-# Decay distance when calculating spreading weights
-FOLD = 15
+# Decay length (rad) when calculating spreading weights
+# spreading should decrease by a factor of e (2.718) every FOLD distance
+FOLD = 2 / 2 * np.pi
```
ezhilsabareesh8: Thanks @anton-seaice. The formula itself is correct — `exp(-d / FOLD)` does decrease by a factor of `e` over one `FOLD` distance. With `FOLD = π` radians, one e-folding distance is about 20,000 km, while the 75 selected ocean cells extend only about 125–345 km from the source. They therefore cover less than 2% of one e-folding distance, so the weights remain almost uniform.

Could you confirm whether `π` radians is intentional? I wondered whether this was meant to represent 2°, for example `np.deg2rad(2)`, which would give an e-folding distance of roughly 220 km.

anton-seaice: oh yeah, deg2rad is clearer thanks

anton-seaice:
Ready to go @ezhilsabareesh8

ezhilsabareesh8 (approved):
Thanks @anton-seaice

anton-seaice:
I made a new set of weights from this, and re-ran ["test4" for 1964 ](https://github.com/ACCESS-NRI/access-om3-configs/issues/1394#issuecomment-4839896985)with MAX_DELTA_SRESTORE = 0.5 and it didn't crash at the same place as test4

Ill let that run out a bit

see /g/data/tm70/as2285/payu/om3_rofl_spread/MC_25km_jra_iaf_spread

anton-seaice:
> Ill let that run out a bit
> 
> see /g/data/tm70/as2285/payu/om3_rofl_spread/MC_25km_jra_iaf_spread

Ran for three years, so seems promising

PR #148, opened 2026-07-31
iceberg pattern with a little smoothing

anton-seaice:
This adds some smoothing to the iceberg melt pattern, and updates to V11 

per https://github.com/ACCESS-NRI/access-om3-configs/issues/404#issuecomment-4849416706

### DIFF
--- a/rof_pattern_generation/generate_rofi_pattern.py
+++ b/rof_pattern_generation/generate_rofi_pattern.py
@@ -30,6 +30,7 @@ import os
 import sys
 from copy import copy
 from sklearn.neighbors import BallTree
+from scipy.ndimage import uniform_filter
 from xesmf.util import cell_area
 import xarray as xr
 import numpy as np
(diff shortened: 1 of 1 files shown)


anton-seaice:
See results in https://github.com/ACCESS-NRI/access-om3-configs/issues/404#issuecomment-5161382731

rof_pattern_generation/generate_rofi_pattern.py
```diff
@@ -189,11 +190,23 @@ def main():
...
-    # After doing the regridding, map any runoff on land cells into the ocean
-    weights_da = move_runoff_on_land(
-        regrid.grid_dest, mask, forcing_regrid_glob["melt"]
+    # apply a spatial smoothing
+    smoothed = forcing_regrid_glob.map(
+        lambda da: xr.apply_ufunc(
+            uniform_filter,
+            da,
+            kwargs={
+                "size": (1, 5, 5),
```
ezhilsabareesh8: is the fixed 5-cell smoothing intended to be resolution independent?

rof_pattern_generation/generate_rofi_pattern.py
```diff
@@ -189,11 +190,23 @@ def main():
...
+                "mode": ("nearest", "nearest", "wrap"),
+            },  # time, y, x
+            dask="parallelized",
+            output_dtypes=[da.dtype],
+        ).clip(min=0)
     )
 
+    # After doing the regridding, map any runoff on land cells into the ocean
+    weights_da = move_runoff_on_land(regrid.grid_dest, mask, smoothed["melt"])
```
ezhilsabareesh8: Is there a reason for applying the smoothing without the ocean mask, before move_runoff_on_land()?

anton-seaice: there's no clear way to smooth with the ocean mask, that I can think of

possibly we could just discard the runoff going to land

rof_pattern_generation/generate_rofi_pattern.py
anton-seaice: ok i put in a fix for this

